PR #4, opened 2017-07-13
Initial analysis schema

dshiga:
The schema for analysis.json won't actually live here. Just making a PR so we can annotate and discuss.


analysis_schema.json
```diff
@@ -0,0 +1,117 @@
...
+      },
+      "analysis_id" : {
+         "description" : "ID give to the analysis run which refers back to the analysis instance in the green box.",
+         "type" : "number"
+      },
+      "metadata_schema" : {
+         "description" : "The version of the metadata schemas used for the json files.",
+         "type" : "string"
+      },
+      "log" : {
```
dshiga: In Cromwell there really is no log of the workflow as a whole. Each step in the workflow has its own stderr and stdout logs. This should be taken out or at least made optional.

analysis_schema.json
```diff
@@ -0,0 +1,117 @@
...
+      "timestamp_stop_utc" : {
+         "type" : "date-time",
+         "description" : "Terminal stop time of the full pipeline."
+      },
+      "sample_json" : {
+         "description" : "The file describing the samples ran.",
+         "type" : "file"
+      },
+      "timings" : {
+         "reference_version" : {
```
dshiga: I would suggest making the reference being just another item in the inputs array and not having its own entry here. I would also suggest pulling inputs and outputs out of timings and make inputs outputs and timings all top level items (siblings) within analysis.json.

analysis_schema.json
```diff
@@ -0,0 +1,117 @@
...
+            "type" : "string",
+            "description" : "Versioned reference used in running the pipeline."
+         },
+         "outputs" : {
+            "type" : "array",
+            "items" : {
+               "type" : {
+                  "$ref" : "#/definitions/timing"
```
dshiga: This looks like a typo. The type for outputs should be "file" not "timing".

analysis_schema.json
```diff
@@ -0,0 +1,117 @@
...
+         "type" : "object",
+         "required" : [
+            "step_name",
+            "step_start",
+            "step_stop"
+         ],
+         "properties" : {
+            "step_name" : "string",
+            "step_stop" : "uri",
+            "step_start" : "uri"
```
dshiga: This should be a date-time, right? Same for step_stop.

analysis_schema.json
```diff
@@ -0,0 +1,117 @@
...
+         "properties" : {
+            "checksum" : "string",
+            "file_path" : "uri",
+            "format" : {
+               "type" : "string",
+               "enum" : [
+                  "text",
+                  "fastq.gz",
+                  "fasta.gz",
+                  "bam"
```
dshiga: Is this an exclusive list? Does it mean we can't represent any other file types with this schema? What about json, cell-gene expression tables, tsv, bai? I worry about trying to maintain a list of file formats that are valid.

analysis_schema.json
```diff
@@ -0,0 +1,117 @@
...
+                  "$ref" : "#/definitions/timing"
+               }
+            },
+            "description" : "Output generatd by the pipeline run."
+         },
+         "inputs" : {
+            "type" : "array",
+            "items" : {
+               "type" : {
+                  "$ref" : "#/definitions/file"
```
dshiga: I think we need to allow a broader set of types here. For example, some inputs may be integer or string parameters needed by steps in the workflow, and not given as files. Other inputs might be arrays or even json objects. Anything that could be passed to Cromwell in the inputs json needs to be possible to capture here, and I believe all types that can be represented in json are accepted by Cromwell as inputs. See list of types here:
https://github.com/broadinstitute/wdl/blob/master/SPEC.md#type-coercion

analysis_schema.json
```diff
@@ -0,0 +1,117 @@
+{
```
dshiga, line 1: Misc thoughts:
-Might be helpful to include input bundle uuid and version somewhere.
-Would it be a good idea to include the docker image and possibly inputs, machine type / memory, for each task?
-We specify "additionalProperties": False a few times below -- does that mean if someone adds more properties to their json document it fails schema validation? Might be nicer to let people add stuff if it's useful to them, unless there's an important reason to forbid it.

analysis_schema.json
```diff
@@ -0,0 +1,117 @@
...
+      },
+      "log" : {
+         "type" : "file",
+         "description" : "A file that is a detailed log of the workflow run."
+      },
+      "timestamp_stop_utc" : {
+         "type" : "date-time",
+         "description" : "Terminal stop time of the full pipeline."
+      },
+      "sample_json" : {
```
dshiga: Why include a pointer to the sample json (and not the various other json files like assay) here?

TimothyTickle: This was originally in to connect the ingest bundle. Current design no longer requires it. removing.

analysis_schema.json
TimothyTickle: We would definitely need a definitive log or tar of logs for a pipeline. Would that not be possible?

analysis_schema.json
TimothyTickle: - Agree on inputs and outputs being top level, typo there. Sorry.
- Separating out reference does have a use case. If you do not separate it out you will need knowledge of the pipeline's parameters to find out which input is the reference. Each pipeline will have a different parameter or transformation of the reference (like a star index) that it will use so you can not be confident by parameter name or data type what the input reference is. Thoughts?

analysis_schema.json
TimothyTickle: Yup, a part of the typo above.

analysis_schema.json
TimothyTickle: I will switch to string, point taken.

analysis_schema.json
dshiga: Yeah, we definitely have logs, but at the task rather than workflow level. I would propose attaching logs to each task in the "timings" section. If we just cat or tar them together it may be difficult for people to figure out which goes with which task.

TimothyTickle: Ok, that works.

analysis_schema.json
TimothyTickle: yup

analysis_schema.json
TimothyTickle: I have this so we have a controlled description of the file types given we can not expect/rely on extensions. We can update as needed. Either way, we need standards on data types and what is to be expected in each task so this information will exist as we develop new pipelines and will need to be controlled in some way so we can rely on it. Happy to hear other ideas.

analysis_schema.json
TimothyTickle: I now have to either string or file.

analysis_schema.json
dshiga: Just to record our offline conversation, we agreed that for convenience of querying we would keep the reference_version as a separate field here (which would be a pointer to a reference bundle in blue box). However, we will also populate the inputs section with a reference parameter, which would most likely be a pointer to a specific file like a fasta.

analysis_schema.json
TimothyTickle: 1. Added and added as a list in case we need to run analysis that extends combines bundles (one of our use cases).
2. Sure, if you feel these will not be parameterized. Which should be added?
3. We will be the only ones writing analysis jsons. With that in mind would you still want to remove the additionalProperties?

dshiga: 2. Okay, I think a few items might be useful especially for analyzing cost and efficiency, listed below:
disk size
docker image
cpus
memory
zone
start_time
stop_time
3. That is a good point. It might still be nice to have flexibility to add additional things without changing the schema -- we would always want to update the schema eventually but perhaps as a stopgap measure to solve some immediate problem. Is there an advantage to locking it down more?

TimothyTickle: 2. Do we want to put these at the top level or be owned by a property?
3. I have opened it up for now at the top level.

dshiga: I think they could be at the top level within each task's "timing" section, i.e. they could all be siblings of step_name. I see you already have step_start and step_stop in there, so I guess those last two in my list are already covered.

TimothyTickle: ok

dshiga:
This looks great, thanks @TimothyTickle! As discussed, I'm going to close this now, since the schema doc is actually going to live in another repo.

PR #18, opened 2017-08-28
Add config for deploying via kubernetes

dshiga:
In order to deploy in Google Container Engine / Kubernetes we need a bunch of configuration files to define the deployment, the service, and the ingress. This PR adds those config files and a script that uses them to set up a container engine cluster and does the deployment.

See https://elastc.com/c/UwfbQgK_/87-prototype-kubernetes-for-mint

### DIFF
--- a/green_box.yml
+++ b/green_box.yml
@@ -6,10 +6,22 @@ info:
 host: localhost:8080
 schemes:
   - https
-basePath: /v1
 produces:
   - application/json
 paths:
+  /:
+    get:
+      summary: health check
+      operationId: 'green_box.api.health.get'
+      responses:
+        200:
+          description: OK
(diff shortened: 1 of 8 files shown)


kbergin:
As part of this PR, or in another, would it be possible to do a little reorganizing of the repo? I'm thinking of the files in the main dir like `mock_provenance.py`, `mock_smartseq2.json`, `mock_smartseq2.wdl`,`notification.json`, `requirements.txt`. I'm not sure whether they belong in this repo or skylab, or if they belong here maybe they could be organized into a test or testdata dir? Thanks :)

dshiga:
Yep, worth doing some cleanup there. I'll do it in a separate PR.

PR #22, opened 2017-09-08
Configuration improvements

dshiga:
Previously, the listener was hard coded to launch the mock smartseq2 wdl. These changes make it configurable, so that we can use the listener to run other wdls. This was necessary in order to test the SmartSeq2 wrapper WDL here:
https://github.com/HumanCellAtlas/skylab/pull/8
Wrapper WDL ticket:
https://elastc.com/c/Kl2fPT8w/177-smartseq2-wrapper-wdl

I also made some improvements here to how we handle configuration in general. The ticket for the config improvements is here: https://elastc.com/c/w3Z04eaS/102-better-way-to-deploy-to-any-environment


ambrosejcarr (approved):
I am guessing that 'test' values in `green_config` were not all `test` when the jenkins test was run, or `test` has some special meaning to cromwell? 

This all looks good to me.

ambrosejcarr:
@dshiga after your clarification about how kubernetes is working and the new changes this is now very transparent. 😄  Everything looks good now to me! :+1:

dshiga:
Forgot to address your other question about the config. The unit tests in Jenkins run with the config in test/config.json, which as you noticed populates everything with the string "test". The unit tests don't actually call Cromwell, so don't need real connection info. I thought about not even providing a config file when running the unit tests, but it's nice to exercise the code that parses the values out of the json.

ambrosejcarr:
Got it. Thanks for the explanation.

PR #23, opened 2017-09-15
Launch workflow depending on the subscription_id of the notification

samanehsan:
Update listener to launch workflow depending on the subscription_id of the notification.

### DIFF
--- a/green_box/api/notifications.py
+++ b/green_box/api/notifications.py
@@ -21,24 +21,40 @@ def post(body):
     logger.info("Notification received")
     logger.info(body)
 
-    # Get bundle uuid and version
-    uuid, version = extract_uuid_version(body)
+    # Get bundle uuid, version and subscription_id
+    uuid, version, subscription_id = extract_uuid_version_subscription_id(body)
+
+    # Find wdl config where the subscription_id matches the notification's subscription_id
+    id_matches = [wdl for wdl in green_config.wdls if wdl.get('subscription_id') == subscription_id]
+    if len(id_matches) == 0:
+        return response_with_server_header(dict(error='Not Found: No wdl config found with subscription id {}'.format(subscription_id)), 404)
+    elif len(id_matches) > 1:
+        return response_with_server_header(dict(error='Internal Server Error: More than one wdl config found with subscription id {}'.format(subscription_id)), 500)
 
     # Prepare inputs
-    inputs = compose_inputs(green_config.workflow_name, uuid, version, green_config.provenance_script)
+    wdl = id_matches[0]
+    workflow_name = wdl.get('workflow_name')
+    inputs = compose_inputs(workflow_name, uuid, version, green_config.provenance_script)
     with open('cromwell_inputs.json', 'w') as f:
         json.dump(inputs, f)
 
     # Start workflow
-    logger.info('Launching {0} workflow in Cromwell'.format(green_config.workflow_name))
-    result = subprocess.check_output(['gsutil', 'cp', green_config.wdl_cloud_path, '.'])
-    result = subprocess.check_output(['gsutil', 'cp', green_config.wdl_deps_cloud_path, '.'])
-    result = subprocess.check_output(['gsutil', 'cp', green_config.default_inputs_cloud_path, '.'])
-    cromwell_response = start_workflow(green_config.wdl_file, green_config.wdl_deps_file, 'cromwell_inputs.json', green_config.default_inputs_file, green_config)
+    logger.info(wdl)
+    logger.info('Launching {0} workflow in Cromwell'.format(workflow_name))
+
+    # Get files from gcs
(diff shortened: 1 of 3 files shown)


dshiga (commented):
This looks great! I have a few suggestions that I'll add.

green_box/api/notifications.py
```diff
@@ -21,24 +21,40 @@ def post(body):
...
+    # Get bundle uuid, version and subscription_id
+    uuid, version, subscription_id = extract_uuid_version_subscription_id(body)
+
+    # launch wdl based on subscription id
+    wdl = None
+    for each in green_config.wdls:
+        if each.get('subscription_id', None) == subscription_id:
+            wdl = each
+    if not wdl:
+        return response_with_server_header(dict(error='Bad Request'), 400)
```
dshiga: I really like that you've added this here. It might be helpful to also include the subscription id in the error message.

green_box/api/notifications.py
```diff
@@ -21,24 +21,40 @@ def post(body):
...
 
-    # Get bundle uuid and version
-    uuid, version = extract_uuid_version(body)
+    # Get bundle uuid, version and subscription_id
+    uuid, version, subscription_id = extract_uuid_version_subscription_id(body)
+
+    # launch wdl based on subscription id
+    wdl = None
+    for each in green_config.wdls:
+        if each.get('subscription_id', None) == subscription_id:
```
dshiga: I like using get here. I would suggest removing the default None argument, since it's the default anyway.

green_box/api/notifications.py
```diff
@@ -21,24 +21,40 @@ def post(body):
...
-    cromwell_response = start_workflow(green_config.wdl_file, green_config.wdl_deps_file, 'cromwell_inputs.json', green_config.default_inputs_file, green_config)
+    logger.info(wdl)
+    logger.info('Launching {0} workflow in Cromwell'.format(workflow_name))
+
+    # Get files from gsc
+    subprocess.check_output(['gsutil', 'cp', wdl['wdl_cloud_path'], '.'])
+    subprocess.check_output(['gsutil', 'cp', wdl['default_inputs_cloud_path'], '.'])
+    subprocess.check_output(['gsutil', 'cp', wdl['wdl_deps_cloud_path'], '.'])
+
+    wdl_file = wdl.get('wdl_file', None)
```
dshiga: It might be better to throw an error at this point if wdl_file or any of the other parameters below are missing. So you could use the wdl_file = wdl['wdl_file'] syntax to achieve that, or else check that all params are populated before calling cromwell below. If you keep the .get calls, I think it's best to not pass the default None argument.

green_box/api/notifications.py
```diff
@@ -21,24 +21,40 @@ def post(body):
...
     # Start workflow
-    logger.info('Launching {0} workflow in Cromwell'.format(green_config.workflow_name))
-    result = subprocess.check_output(['gsutil', 'cp', green_config.wdl_cloud_path, '.'])
-    result = subprocess.check_output(['gsutil', 'cp', green_config.wdl_deps_cloud_path, '.'])
-    result = subprocess.check_output(['gsutil', 'cp', green_config.default_inputs_cloud_path, '.'])
-    cromwell_response = start_workflow(green_config.wdl_file, green_config.wdl_deps_file, 'cromwell_inputs.json', green_config.default_inputs_file, green_config)
+    logger.info(wdl)
+    logger.info('Launching {0} workflow in Cromwell'.format(workflow_name))
+
+    # Get files from gsc
```
dshiga: should be gcs

samanehsan:
Thanks for the comments @dshiga! I updated the PR with those changes :octocat:

green_box/api/notifications.py
```diff
@@ -21,24 +21,40 @@ def post(body):
     logger.info("Notification received")
     logger.info(body)
 
-    # Get bundle uuid and version
-    uuid, version = extract_uuid_version(body)
+    # Get bundle uuid, version and subscription_id
+    uuid, version, subscription_id = extract_uuid_version_subscription_id(body)
+
+    # launch wdl based on subscription id
+    wdl = None
```
ambrosejcarr: If I'm reading this pattern correctly, you're testing for the lack of a matching wdl response. I think it might be a bit clearer if written as follows: 

```
id_matches = [wdl.get('subscription_id') == subscription_id for wdl in green_config.wdls]
if not any(id_matches):
    return response_with_server_header(dict(error='Bad Request: Invalid subscription id {}'.format(subscription_id)), 400)
```

ambrosejcarr (commented):
Great job - one minor comment about one section of the code I had to re-read. Please forgive if I didn't understand it correctly. :)

samanehsan:
Thanks for the suggestions @ambrosejcarr! Let me know if those changes look good :octocat:

ambrosejcarr (approved):
Looks great!

dshiga (approved):
Looks great!

PR #24, opened 2017-09-20
Update registered query to match v3 metadata

samanehsan:
- Update smartseq2-query.json and add 10X-query.json to match v3 metadata schema.
- Make `subscriptions_list.py` take in the `key_file` as a parameter
- Update `requirements.txt`


scripts/config.sh
```diff
@@ -2,3 +2,4 @@ dss_url=test
 green_url=test
 key_file=test
 listener_secret=test
+query_json=smartseq2-query.json
```
ambrosejcarr: Github wants you to add a newline at end of file 👼 .

scripts/subscription-create.py
```diff
@@ -31,6 +31,7 @@ def make_request(js, dss_url, key_file):
     parser.add_argument("callback_base_url")
     parser.add_argument("listener_secret")
     parser.add_argument("key_file")
+    parser.add_argument("query_json")
```
ambrosejcarr: long-term: (I know this isn't in scope of PR: these arguments could use a `help='description'` and `required=True` field)

ambrosejcarr (approved):
World's most minor comments. LGTM.

scripts/subscription-create.py
jishuxu: Yes, agree with Ambrose on this.

scripts/config.sh
jishuxu: should you add " #!/bin/bash" at the beginning?

jishuxu: Also 'test' is string or variable? It might throw out error if it is not clearly defined?

requirements.txt
```diff
@@ -1 +1,4 @@
+argparse
 connexion
+httplib2
+oauth2client
```
dshiga: It's important to keep the requirements for running the listener separate from those needed to run the subscription scripts.

I'd suggest the following:
Delete this requirements.txt. Update the Dockerfile to pip install connexion, rather than installing from requirements.txt. We will always run the listener via docker and having a requirements.txt for it is just confusing I think. Then add a new requirements.txt in the scripts directory. That one would just contain the requirements for the scripts themselves, which should just be httplib2 and oauth2client (argparse shouldn't need installing, should just be available).

scripts/10X-query.json
```diff
@@ -0,0 +1,19 @@
...
+          "files.assay_json.rna.library_construction": "10x_v2"
+        }
+      }, {
+        "match": {
+          "files.sample_json.donor.species.ontology": 9606
+        }
+      }]
+    }
+  }
+}
```
dshiga: Could you move these query json files to a v3 directory? The metadata schema is going to keep evolving and this will help clarify which files go with which schema version.

scripts/subscription-create.py
dshiga: That's a good suggestion. I don't think it's out of scope.

scripts/config.sh
dshiga: Good points. Could you add a README.md with a very brief description of what the files in the scripts directory are for? I think that would reduce confusion. Maybe rename config.sh to config-template.sh and explain that you need a file called config.sh with real values in it. The #! is not needed in config-template.sh because it will never be run (and config.sh only gets sourced in, which means the #! would get ignored anyway). Lastly, could you remove query_json from this file and instead require users of subscription-create.sh to call it like:  ./subscriptions-create.sh smartseq2-query.json
That way we don't need to edit any files in order to change which json file we supply.

scripts/subscription-create.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
 source config.sh
-./subscription-create.py "$dss_url?replica=aws" $green_url $listener_secret $key_file
+./subscription-create.py "$dss_url?replica=gcp" $green_url $listener_secret $key_file $query_json
```
dshiga: To follow up on my comment about config.sh, I'd suggest changing this file to expect the user to pass in the path to the query json file as an argument.

scripts/subscriptions-list.py
```diff
@@ -4,17 +4,18 @@
...
 from httplib2 import Http
 from oauth2client.service_account import ServiceAccountCredentials
-from apiclient.discovery import build
 
-def run(dss_url):
+
+def run(dss_url, key_file):
     scopes = ['https://www.googleapis.com/auth/userinfo.email']
-    credentials = ServiceAccountCredentials.from_json_keyfile_name('../../green-box-config/dev/bluebox-subscription-manager.json', scopes)
+    credentials = ServiceAccountCredentials.from_json_keyfile_name(key_file, scopes)
```
dshiga, line 11: Ahh, this is much better, thanks!

dshiga (commented):
Looks great! Thanks for all your improvements to these scripts. Could you please also try out the subscription delete script and fix problems there? I think some of the fixes you've made here are also needed there. You could experiment in dev to avoid disrupting the staging environment.

scripts/subscription-create.py
samanehsan: Good call on adding `help` text! Since these are positional arguments they are required (and I actually get an error if I specify `required=True`)

ambrosejcarr: Oops, sorry! I forgot about the `--/-` prefix distinction for position/optional. You're completely right.

Dockerfile
```diff
@@ -14,7 +14,7 @@ WORKDIR /secondary-analysis
 COPY requirements.txt .
 
 RUN pip install wheel && \
-  pip install -r requirements.txt
+  pip install connexion
```
dshiga: can consolidate a bit here, like this::

    RUN pip install \
       wheel \
       connection

scripts/README.md
```diff
@@ -0,0 +1,13 @@
+###Set-up
```
dshiga, line 1: This is great, thanks for adding this!

dshiga (approved):
Great! Please tweak the pip install command as described in comment on Dockerfile, then I think this is ready to merge!

PR #26, opened 2017-09-26
Don't trigger on analysis bundles

dshiga:
Small changes to queries:
-Exclude analysis bundles
-Tweak json formatting
-Correct FACs to FACS
-Rename 10X-query.json to 10x-query.json

See https://elastc.com/c/Sf8i*3IW/235-avoid-infinite-loop-in-listener-processing


samanehsan (approved):
Looks good! :sunny:

PR #27, opened 2017-09-27
Provide Cromwell options file

dshiga:
Some Cromwell behavior can be configured on a per workflow basis, by providing an options json file along with the wdl and inputs. Currently, however, the listener does not pass along an inputs file when launching workflows. This PR allows (requires) an options file to be provided along with the wdl and inputs for each workflow.

See https://elastc.com/c/kVFnsQYu/234-enable-listener-to-launch-without-call-caching

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,9 +11,9 @@ RUN apt-get -y install --no-install-recommends \
 RUN mkdir /secondary-analysis
 WORKDIR /secondary-analysis
 
-RUN pip install \
-  wheel \
-  connexion
+RUN pip install wheel
+
+RUN pip install connexion
 
 RUN apt-get install -y lsb-release curl && \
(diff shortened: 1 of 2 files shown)


rexwangcc:
The code looks good to me based on what I've read below:
1. https://github.com/broadinstitute/cromwell#call-caching
2. https://github.com/broadinstitute/cromwell#workflow-options

We would figure out how to design the unittest of this feature later after other changes are done!

rexwangcc (approved):
1. For dockerfile fix: Splitting one line `pip install` commands back into multiple lines so the installation of those libraries depend on wheel won't fail. Looks great to me!

2. For listener option feature: The code looks good to me based on what I've read below:
    * https://github.com/broadinstitute/cromwell#call-caching
    * https://github.com/broadinstitute/cromwell#workflow-options

And we would figure out how to design the unittest of this feature later after other changes are done!

PR #29, opened 2017-09-28
Ac listener startup checks

ambrosejcarr:
This PR introduces a `Config` class that has methods to check the correctness of `config.json`

#### This `Config` class is subclassed twice to do specialized checks: 
- `WdlConfig` to check wdl specifications
- `ListenerConfig` to check the config.json top-level keys

#### Current tests for `config.json`:
- checks for any missing required keys
- checks for any unexpected keys
- checks for duplicated keys
- checks for duplicated wdl definitions
- checks for multiple wdl definitions with the same subscription id

#### Implementation Notes
1. As of now, the `unittests` are all contained in the same file; because of the way `green_box.__init__` is configured to run code, normal separation of testing files is tricky. 
2. By switching to a `Config` class, I was able to simplify a lot of the calls made in `notifications.py`. There is a small amount of additional code that was required use this class as a drop-in replacement for the dictionary-like object expected by the `connexion` app. 
3. Please note the "true" config file that is not stored in this repository will need to be reflect the simplification of the wdl specification in `config.json`. Currently there are both `<arg>_path` and `<arg>_file`, which has been simplified to `<arg>_link` with the file automatically inferred. 
4. In the process of testing this, I realized I was able to break `notifications.py` in many ways that are not caught by the `test.sh` script; we should probably make those tests more robust.

### DIFF
--- a/green_box/api/notifications.py
+++ b/green_box/api/notifications.py
@@ -5,15 +5,18 @@ import logging
 import json
 import subprocess
 import time
-from collections import namedtuple, OrderedDict
 from flask import make_response, current_app
+
+
+def get_filename_from_gs_link(link):
+    """get the filename corresponding to a google_storage link"""
+    return link.split('/')[-1]
+
 
 def post(body):
     # Check authentication
     logger = logging.getLogger('green-box')
-    ordered_config = OrderedDict(current_app.config)
-    Config = namedtuple('config', ordered_config.keys())
-    green_config = Config(*ordered_config.values())
+    green_config = current_app.config
     if not is_authenticated(connexion.request.args, green_config.notification_token):
         time.sleep(1)
         return response_with_server_header(dict(error='Unauthorized'), 401)
@@ -25,35 +28,37 @@ def post(body):
     uuid, version, subscription_id = extract_uuid_version_subscription_id(body)
 
     # Find wdl config where the subscription_id matches the notification's subscription_id
-    id_matches = [wdl for wdl in green_config.wdls if wdl.get('subscription_id') == subscription_id]
+    id_matches = [wdl for wdl in green_config.wdls if wdl.subscription_id == subscription_id]
     if len(id_matches) == 0:
-        return response_with_server_header(dict(error='Not Found: No wdl config found with subscription id {}'.format(subscription_id)), 404)
-    elif len(id_matches) > 1:
-        return response_with_server_header(dict(error='Internal Server Error: More than one wdl config found with subscription id {}'.format(subscription_id)), 500)
+        return response_with_server_header(
+            dict(error='Not Found: No wdl config found with subscription id {}'
+                       ''.format(subscription_id)), 404)
+    wdl = id_matches[0]
 
     # Prepare inputs
-    wdl = id_matches[0]
-    workflow_name = wdl.get('workflow_name')
-    inputs = compose_inputs(workflow_name, uuid, version, green_config.provenance_script)
+    inputs = compose_inputs(wdl.workflow_name, uuid, version)
     with open('cromwell_inputs.json', 'w') as f:
         json.dump(inputs, f)
 
     # Start workflow
     logger.info(wdl)
-    logger.info('Launching {0} workflow in Cromwell'.format(workflow_name))
+    logger.info('Launching {0} workflow in Cromwell'.format(wdl.workflow_name))
 
-    # Get files from gcs
-    subprocess.check_output(['gsutil', 'cp', wdl['wdl_cloud_path'], '.'])
-    subprocess.check_output(['gsutil', 'cp', wdl['default_inputs_cloud_path'], '.'])
-    subprocess.check_output(['gsutil', 'cp', wdl['wdl_deps_cloud_path'], '.'])
-    subprocess.check_output(['gsutil', 'cp', wdl['options_cloud_path'], '.'])
+    # Get files from gcs # todo change to using google.cloud.storage
+    subprocess.check_output(['gsutil', 'cp', wdl.wdl_link, '.'])
+    subprocess.check_output(['gsutil', 'cp', wdl.wdl_default_inputs_link, '.'])
+    subprocess.check_output(['gsutil', 'cp', wdl.wdl_deps_link, '.'])
+    subprocess.check_output(['gsutil', 'cp', wdl.options_link, '.'])
 
-    wdl_file = wdl['wdl_file']
-    wdl_deps_file = wdl['wdl_deps_file']
-    wdl_default_inputs_file = wdl['default_inputs_file']
-    options_file = wdl['options_file']
-    cromwell_response = start_workflow(wdl_file, wdl_deps_file, 'cromwell_inputs.json',
-                                        wdl_default_inputs_file, options_file, green_config)
+    # get filenames from links
+    wdl_file = get_filename_from_gs_link(wdl.wdl_link)
+    wdl_default_inputs_file = get_filename_from_gs_link(wdl.wdl_default_inputs_link)
+    wdl_deps_file = get_filename_from_gs_link(wdl.wdl_deps_link)
+    options_file = get_filename_from_gs_link(wdl.options_link)
+
+    cromwell_response = start_workflow(
+        wdl_file, wdl_deps_file, 'cromwell_inputs.json',
+        wdl_default_inputs_file, options_file, green_config)
 
     # Respond
     if cromwell_response.status_code > 201:
@@ -62,18 +67,21 @@ def post(body):
     logger.info(cromwell_response.json())
     return response_with_server_header(cromwell_response.json())
 
+
 def response_with_server_header(body, status=200):
     response = make_response(json.dumps(body), status)
     response.headers['Server'] = 'Secondary Analysis Service'
     response.headers['Content-type'] = 'application/json'
     return response
 
+
 def is_authenticated(args, token):
-    if not 'auth' in args:
+    if 'auth' not in args:
         return False
     elif args['auth'] != token:
         return False
     return True
+
 
 def extract_uuid_version_subscription_id(msg):
     uuid = msg["match"]["bundle_uuid"]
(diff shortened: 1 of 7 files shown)


green_box/api/notifications.py
```diff
@@ -25,35 +23,32 @@ def post(body):
...
+            dict(error='Not Found: No wdl config found with subscription id {}'
+                       ''.format(subscription_id)), 404)
+    wdl = id_matches[0]
 
     # Prepare inputs
-    wdl = id_matches[0]
-    workflow_name = wdl.get('workflow_name')
-    inputs = compose_inputs(workflow_name, uuid, version, green_config.provenance_script)
+    inputs = compose_inputs(
+        wdl.workflow_name, uuid, version, green_config.provenance_script)
```
dshiga: The provenance_script is a holdover from when we were running the mock smartseq2 workflow via the listener. Could you remove it? None of the workflows should need it as a separate input now.

green_box/api/notifications.py
```diff
@@ -25,35 +23,32 @@ def post(body):
...
     logger.info(wdl)
-    logger.info('Launching {0} workflow in Cromwell'.format(workflow_name))
+    logger.info('Launching {0} workflow in Cromwell'.format(wdl.workflow_name))
 
-    # Get files from gcs
-    subprocess.check_output(['gsutil', 'cp', wdl['wdl_cloud_path'], '.'])
-    subprocess.check_output(['gsutil', 'cp', wdl['default_inputs_cloud_path'], '.'])
-    subprocess.check_output(['gsutil', 'cp', wdl['wdl_deps_cloud_path'], '.'])
-    subprocess.check_output(['gsutil', 'cp', wdl['options_cloud_path'], '.'])
+    # Get files from gcs # todo change to using google.cloud.storage
```
dshiga: Thanks for adding this

green_box/api/notifications.py
ambrosejcarr: Just to be clear, `provenance_script` is an input for the ListenerConfig, not the individual wdls (it is not repeated, but is present only once in config at the same level as, for example, the cromwell password). 

Are you saying that `provenance_script` is not needed at all anywhere? if so, will remove.

dshiga: Yeah, we shouldn't need it anywhere anymore. It was just a silly way of adding an extra input needed by the mock smartseq2 workflow. We're not really running that workflow anymore and even if we do, we can just add that input to the default inputs json, rather than having a whole separate config entry for it.

green_box/validate_config.py
```diff
@@ -0,0 +1,269 @@
...
+
+        wdl_keys = set(self.config_dictionary)
+        extra_keys = wdl_keys - self.expected_fields
+        missing_keys = self.expected_fields - wdl_keys
+        if missing_keys:
+            raise ValueError(
+                'The following WDL configuration is missing key(s): {keys}\n{wdl}'
+                ''.format(wdl=self.config_dictionary, keys=', '.join(missing_keys)))
+        if extra_keys:
+            raise ValueError(
```
dshiga: Could we log a warning here instead of raising an error? A warning will hopefully provide enough protection while still giving us flexibility to add config entries without a code change, should we ever need it. We could then rename expected_fields to required_fields.

green_box/validate_config.py
```diff
@@ -0,0 +1,269 @@
...
+                'The following WDL configuration has unexpected key(s): {keys}\n{wdl}'
+                ''.format(wdl=self.config_dictionary, keys=', '.join(extra_keys)))
+
+    def __eq__(self, other):
+        if isinstance(other, WdlConfig) and hash(self) == hash(other):
```
dshiga: how about return isinstance(other, WdlConfig) and hash(self) == hash(other)?

green_box/validate_config.py
```diff
@@ -0,0 +1,269 @@
...
+            elif isinstance(field_value, Config):
+                result += field_value.to_string()
+            else:
+                raise ValueError('Config fields must either be strings or config objects')
+        return result
+
+    def __hash__(self):
+        return hash(self.to_string())
+
+    # needed for flask interface
```
dshiga: helpful comment, thanks

green_box/validate_config.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+class WdlConfig(Config):
+    """subclass of Config to check WDL configurations"""
+
+    def __init__(self, json_config_object, *args):
+        Config.__init__(self, json_config_object, *args)
+
+        # add filename information for google-storage links in the config file
+        for k, v in self.config_dictionary.items():
```
dshiga: I think it might be more concise to move the _get_filename_from_gs_link function into notifications.py and just call it everywhere we need to get a file from a path.

ambrosejcarr: Great comment, removed a lot of code. 👍

green_box/validate_config.py
ambrosejcarr: Yep, that is much better.

green_box/validate_config.py
ambrosejcarr: Done.

green_box/api/notifications.py
ambrosejcarr: 👍

green_box/api/notifications.py
ambrosejcarr: Ok; removed. Please note that I needed to modify one of the tests (as now `compose_inputs` takes one fewer argument).

ambrosejcarr:
Thanks for the comments! back to you @dshiga

green_box/validate_config.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def test_listener_config_exposes_all_methods_requested_in_notifications(self):
+        # test that the call smade in notifications refer to attributes that
+        # ListenerConfig exposes
+        config = ListenerConfig(correct_test_config)
+        requested_listener_attributes = [
+            'notification_token', 'wdls', 'cromwell_url', 'cromwell_user',
+            'cromwell_password']
+        for attr in requested_listener_attributes:
+            self.assertTrue(hasattr(config, attr), 'missing attribute %s' % attr)
```
dshiga: I didn't realize you could add a custom error message here - I like it!

green_box/validate_config.py
```diff
@@ -0,0 +1,238 @@
...
+            'wdl_deps_link',
+            'wdl_default_inputs_link',
+            'options_link'
+        }
+
+
+class ListenerConfig(Config):
+    """subclass of Config to check listener configurations"""
+
+    def __init__(self, json_config_object, *args):
```
dshiga: What does the *args do here? It doesn't seem like we use it.

green_box/validate_config.py
```diff
@@ -0,0 +1,238 @@
...
+""")
+
+
+class TestStartupVerification(unittest.TestCase):
+
+    def test_correct_config_throws_no_errors(self):
+        """No errors should be generated"""
+        config = deepcopy(correct_test_config)
+        result = ListenerConfig(config)
+        self.assertIsInstance(result, ListenerConfig)
```
dshiga: How about doing the try, except, self.fail thing like here?
https://stackoverflow.com/questions/4319825/python-unittest-opposite-of-assertraises

That would give us more control over the error message. If an error is thrown in ListenerConfig init, we'll never reach the assertIsInstance.

green_box/validate_config.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def test_config_missing_any_field_throws_value_error(self):
```
dshiga: Maybe we should call this ...missing_required_field... instead of ...missing_any_field.., now that the check is for required fields only.

green_box/validate_config.py
```diff
@@ -0,0 +1,238 @@
...
+
+        def delete_wdl_field(field_name):
```
dshiga: This is great, really like that we're checking this in Config init now and testing it here.

green_box/validate_config.py
```diff
@@ -0,0 +1,238 @@
...
+                'One or more wdl specifications contains a duplicated subscription ID '
+                'but have non-identical configurations. Please check configuration file '
+                'contents.')
+
+
+"""
+################################### TESTS ###########################################
+Because green_box.__init__.py runs code in its body, relative importing of files
+downstream of that package is broken. Thus, the tests are included in this module and
+can be run by invoking `python2.7 validate_config.py`
```
dshiga: Thanks for this helpful explanation. Is it possible to refactor a little bit to make it possible to import the way we import green_box.api in test_service.py? I think that would mean making validate_config a directory with its own __init__.py and validate_config.py in it (or maybe more accurate to just call it config/config.py).

green_box/validate_config.py
```diff
@@ -0,0 +1,238 @@
...
+      "subscription_id": 1,
+      "wdl_link": "gs://broad-dsde-mint-dev-teststorage/wrapper_ss2_single_sample.wdl",
+      "workflow_name": "WrapperSs2RsemSingleSample",
+      "wdl_deps_link": "gs://broad-dsde-mint-dev-teststorage/ss2_single_sample_analysis_only.zip",
+      "wdl_default_inputs_link": "gs://broad-dsde-mint-dev-teststorage/wrapper_ss2_single_sample_example_static.json",
+      "options_link": "gs://broad-dsde-mint-dev/an_options_file.json"
+    }
+  ]
+}
+""")
```
dshiga: I'd be inclined to move this out into its own file in a test/data dir, and reading it in for tests similar to this:
https://github.com/HumanCellAtlas/skylab/blob/ds_ss2_submission_76/docker/secondary-analysis-python/tools/submit/tests/test_create_analysis_json.py

It feels a bit cluttery to have the test data inline here, but on the other hand, in some ways it's convenient to have it here, so I could be persuaded otherwise.

green_box/validate_config.py
```diff
@@ -0,0 +1,238 @@
...
+            for wdl in json_config_object['wdls']:
+                wdl_configs.append(WdlConfig(wdl))
+        except KeyError:
+            raise ValueError('supplied config file must contain a "wdls" section.')
+        self._verify_wdl_configs(wdl_configs)
+        json_config_object['wdls'] = wdl_configs
+
+        Config.__init__(self, json_config_object, *args)
+
+        self.MAX_CONTENT_LENGTH = 10 * 1000
```
dshiga: While you're revamping config, is there a way we can expose this field to make it configurable in the config.json? If it requires a lot of trouble it's probably not worth it, but if it's easy to do it would be nice to be able to tweak this without a code change.

dshiga (commented):
This is great! I feel like we're getting a good bargain now where we're protecting ourselves from silly config mistakes with only a small amount of new code. I appreciate all the documentation you've put in the code and the PR, setting a good example for us to follow! I made a few suggestions in comments.

green_box/validate_config.py
ambrosejcarr: It's a catch-all for arguments that are passed unmodified to the super class's (`Config's`) constructor. In this case, it can receive the flask dictionary. I should also add `kwargs` in case the user passes it as a named argument, which in the current form would break the subclass constructor.

ambrosejcarr:
Ok, changes addressed. To do the testing refactor I had to give up on local testing, which I think is probably ok; now to test you build the docker and test with `test/test.sh`. As a corollary, Jenkins will now run my tests as well.

dshiga (approved):
This is really great! I think it's a huge improvement over what we were doing for config before. Would you mind renaming validate_config.py to just config.py? We are doing validation in there but also defining the config classes themselves and if I weren't familiar with the repo I would probably go looking for something named more like config.py for the config classes themselves. Other than that I think this is ready to merge.

PR #33, opened 2017-11-07
Draft: Create integration test

dshiga:
This is very rough, but wanted to get some early feedback.

Adds integration_test.sh, which builds the listener, renders its config, starts it, sends in notifications and waits for the workflows to complete.

Also modified green_box_utils to install some smaller google cloud libraries instead of the one big google cloud library (which has more dependencies and was causing the build to break).

Some of the comments may be out of date, will fix.

I'm looking for feedback on the overall approach and any issues it raises. The challenge here is that integration testing requires getting a bunch of components working together, which requires a lot of parameters. We have to get those parameters from somewhere, but exactly where we should best store all this info (in mint-deployment repo tsvs, various config files, vault, etc), and how we retrieve and use it is a bit up in the air.

Adding to the challenge is that we want to be able to trigger this integration test script in multiple ways, which may require us to gather the parameters in different ways. Ideally, this one test script will get triggered by PRs in secondary-analysis, skylab, and pipeline-tools, and will also be runnable locally by devs. Each of those scenarios involves gathering the parameters needed in the script in slightly different ways. The dev_config.json, dev_secrets.json, and configtemplate.json files would need to be stored somewhere where they could be retrieved by a CI job or generated.

The script assumes that wdls have already been deployed. See wdl deploy script here:
https://github.com/HumanCellAtlas/skylab/pull/40

We want this script to be simple and robust, while providing the flexibility needed to cover all of these scenarios.

### DIFF
--- a/green_box_utils/green_box_utils/gcs_utils.py
+++ b/green_box_utils/green_box_utils/gcs_utils.py
@@ -2,7 +2,7 @@
 """
 import logging
 from google.cloud import storage
-from google.oauth2 import service_account
+import google.auth
 from io import BytesIO
 
 
@@ -24,9 +24,10 @@ def parse_bucket_blob_from_gs_link(path):
     """
     if not path.startswith('gs://'):
         raise ValueError('%s path is not a valid link')
-    (prefix, _, bucket), blob = path.split('/')[:3], path.split('/')[3:]
+    parts = path.split('/', 3)
+    (prefix, _, bucket), blob = parts[:3], parts[3]
 
-    return bucket, '/'.join(blob)
+    return bucket, blob
 
 
 def download_to_buffer(blob):
@@ -59,10 +60,11 @@ def download_gcs_blob(gcs_client, bucket_name, source_blob_name):
 
     bucket = authenticated_gcs_client.bucket(bucket_name)
     blob = bucket.blob(source_blob_name)
+    logging.debug('bucket: {0}, blob: {1}'.format(bucket_name, source_blob_name))
     return download_to_buffer(blob)
 
 
-class LazyProperty:
+class LazyProperty(object):
     """This class implements a decorator for lazy-initializing class properties.
 
         Instead of implementing Singleton Pattern, this decorator accepts multiple
@@ -80,9 +82,10 @@ class LazyProperty:
         else:
             val = self.func(instance)
             setattr(instance, self.func.__name__, val)
+            return val
 
 
-class GoogleCloudStorageClient:
+class GoogleCloudStorageClient(object):
     def __init__(self, key_location, scopes):
         """This class implements the client to interact with Google Cloud Storage.
 
@@ -99,7 +102,7 @@ class GoogleCloudStorageClient:
         """
         logging.getLogger()
         logging.debug('Configuring listener credentials using %s' % self.key_location)
-        credentials = service_account.Credentials.from_service_account_file(
-            self.key_location, scopes=self.scopes)
-        client = storage.Client(credentials=credentials, project=credentials.project_id)
+
+        credentials, project = google.auth.default(scopes=self.scopes)
+        client = storage.Client(credentials=credentials, project=project)
         return client
--- a/green_box_utils/setup.py
+++ b/green_box_utils/setup.py
@@ -21,7 +21,8 @@ setup(
     install_requires=['Flask>=0.12.2',
                       'connexion>=1.1.15',
                       'requests>=2.18.4',
-                      'google-cloud',
+                      'google-auth',
+                      'google-cloud-storage',
                       'requests-mock',
                       'mock'],
 )
--- a/listener-start.sh
+++ b/listener-start.sh
@@ -1,6 +1,3 @@
 #!/usr/bin/env bash
 
-gcloud config set project broad-dsde-mint-dev
-gcloud auth activate-service-account --key-file=/etc/secondary-analysis/bucket-reader-key.json
-
 python green-box-api.py
--- a/test/get_latest_release.py
+++ b/test/get_latest_release.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python
+
+import argparse
+import requests
+import re
+
+class Tag:
+    def __init__(self, major, minor, patch):
+        self.v = [major, minor, patch]
+        self.string = '{0}.{1}.{2}'.format(self.v[0], self.v[1], self.v[2])
+
+    def __str__(self):
+        return self.string
+
+    def __eq__(self, other):
+        return self.string == other.string
+
+    def __gt__(self, other):
+        for s, o in zip(self.v, other.v):
+            if s < o:
+                return False
+            if s > o:
+                return True
+        return False
+
+def run(repo):
+    url = 'https://api.github.com/repos/{0}/git/refs/tags/'.format(repo) 
+    response = requests.get(url)
+    if response.status_code != 200:
+        raise ValueError('Return code was {0} for {1}, message:\n'.format(response.status_code, url, response.json()))
+    
+    p = re.compile(r'^(\d+)\.(\d+)\.(\d+)$')
+    tags = []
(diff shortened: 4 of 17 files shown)


green_box_utils/green_box_utils/gcs_utils.py
```diff
@@ -2,7 +2,7 @@
 """
 import logging
 from google.cloud import storage
-from google.oauth2 import service_account
+from oauth2client import service_account
```
rexwangcc: I'm not sure if this change could work, since `dir(service_account)` doesn't provide `Credentials`, which is required by `class GoogleCloudStorageClient`, maybe we also need to modify the class as well?

green_box_utils/setup.py
```diff
@@ -21,7 +21,8 @@
     install_requires=['Flask>=0.12.2',
                       'connexion>=1.1.15',
                       'requests>=2.18.4',
-                      'google-cloud',
+                      'google-api-python-client',
+                      'google-cloud-storage',
```
rexwangcc: This change is really neat! Also, as @dshiga and I discussed offline, it would be great if we freeze the version of the dependencies.

test/10x_notification.json
```diff
@@ -0,0 +1,9 @@
+{
+  "transaction_id": "eed1414b-4e70-45a3-b6d1-7ec19ce5b666",
+  "subscription_id": "3e3e176b-629f-46ea-b01d-e36bd650dc54",
+  "match": {
```
rexwangcc: This is not secret, is it?

test/render_lira_config.py
```diff
@@ -0,0 +1,33 @@
...
+        env_config = json.load(f)
+    with open(args.secrets_file) as f:
+        secrets_config = json.load(f)
+
+    with open(args.template_file) as f:
+        s = f.read()
+
+    params_dict = vars(args)
+    params_dict.update(env_config)
+    for k, v in params_dict.iteritems():
```
rexwangcc: Maybe I'd use `for k, v in params_dict.items():` here to provide py2/3 compatibility, although `items()` in python2 would be a little bit more expensive than `iteritems()`.

test/10x_notification.json
dshiga: No, not secret

rexwangcc: Thanks, it looks like a string with a high entropy, sorry for that, ignore me.

dshiga: I'm glad you're keeping an eye on these things!

test/integration_test.sh
```diff
@@ -0,0 +1,180 @@
...
+# env=int
+# lira_version=pipeline-tools PR branch name if exists else latest release
+# pipeline_tools_version=pipeline-tools PR branch name
+# skylab_version = latest versions deployed to int
+
+set -e
+
+env=$1
+mint_deployment_dir=$2
+lira_mode=$3
```
samanehsan, line 71: I think a comment about what the mode parameter means and what options are available for `lira_mode`, `infra_mode`, `tenx_mode` and `ss2_mode` would be useful here

test/current_deployed_version.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python
+
+import argparse
+import csv
+
+def run(pipeline_name, mint_deployment_dir, env):
```
samanehsan: I think this script could use a more specific name like `current_deployed_pipeline_version.py`, as well as a docstring describing what it's doing

test/get_latest_release.py
```diff
@@ -0,0 +1,48 @@
...
+    def __gt__(self, other):
+        for s, o in zip(self.v, other.v):
+            if s < o:
+                return False
+            if s > o:
+                return True
+        return False
+
+def run(repo):
+    url = 'https://api.github.com/repos/{0}/git/refs/tags/'.format(repo)
```
samanehsan, line 27: Are these tags not returned in a particular order?

test/integration_test.sh
```diff
@@ -0,0 +1,180 @@
...
+source integration-test-env/bin/activate
+pip install requests
+tenx_workflow_id=$(python $script_dir/send_notification.py \
+                  --lira_url "http://localhost:8080/notifications" \
+                  --secrets_file dev_secrets.json \
+                  --notification $lira_dir/test/10x_notification.json)
+ss2_workflow_id=$(python $script_dir/send_notification.py \
+                  --lira_url "http://localhost:8080/notifications" \
+                  --secrets_file dev_secrets.json \
+                  --notification $lira_dir/test/ss2_notification.json)
```
samanehsan: The subscription id and possibly the bundle id in a notification will be different for dev. vs. staging. How about including notification files for both dev and staging and changing this to `ss2_notification_$env.json`?

test/integration_test.sh
```diff
@@ -0,0 +1,180 @@
...
+fi
+
+# 5. Render config.json
+# (TODO: Use Henry's script here)
+# TODO: use config file from config repo
+# dev_secrets.json will come from Vault eventually
+echo "Rendering Lira config"
+python $script_dir/render_lira_config.py \
+    --template_file configtemplate.json \
+    --env_config_file dev_config.json \
```
samanehsan: The `env_config_file` and `secrets_file` that get passed in should change depending on the environment

listener-start.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-gcloud config set project broad-dsde-mint-dev
-gcloud auth activate-service-account --key-file=/etc/secondary-analysis/bucket-reader-key.json
+#gcloud config set project broad-dsde-mint-dev
```
samanehsan: Can these lines be removed?

dshiga: Yes, will do

test/render_lira_config.py
dshiga: Good idea!

test/get_latest_release.py
dshiga: They seem to be returned in an order that puts the highest version last, but I haven't been able to find documentation to that effect, so would rather not rely on it.

test/integration_test.sh
dshiga: Good point, will do

test/integration_test.sh
dshiga: Good point

test/current_deployed_version.py
dshiga: Okay, will do

test/integration_test.sh
dshiga: good idea, will do

test/integration_test.sh
```diff
@@ -0,0 +1,179 @@
...
+# The lira_mode param can be "local", "image" or "github".
+# If "local" is specified, a local copy of the Lira code is used and
+# lira_version is ignored. If "image" is specified, this script will pull and run
+# a particular version of the Lira docker image specified by lira_version.
+# Running in "github" mode is not fully implemented yet, but is intended to clone
+# the Lira repo and check out a specific branch, tag, or commit to use, specified
+# by lira_version.
+#
+# pipeline_tools_mode and pipeline_tools_version
+# When pipeline_tools_mode is "current", this script will find the currently deployed
```
samanehsan: It looks like the mode variable for lira specifies where to get the lira code from, but for `pipeline_tools_mode` "current" is referring to which version to use. Could we set `pipeline_tools_version` to "current" instead?

test/integration_test.sh
samanehsan: Thanks for adding more documentation!

test/get_latest_release.py
samanehsan: Oh okay! Could you add in a comment explaining that as well?

test/integration_test.sh
dshiga: Good point - I added a commit to address this question.

samanehsan (approved):
Looks good to me!

PR #35, opened 2017-11-17
Remove green_box_utils

dshiga:
As part of repo reorganization, I'd like to get rid of the green_box_utils package inside lira and distribute its parts elsewhere:

* cromwell-tools repo
for start_workflow, which is a generic function for starting workflows in Cromwell
https://github.com/broadinstitute/cromwell-tools/pull/2

* lira/utils.py
for functions that are useful only inside lira

* pipeline-tools repo
for gcs_utils, which is a set of generic functions for working with gcs and may be useful elsewhere
https://github.com/HumanCellAtlas/pipeline-tools/pull/3

I took the opportunity to reorganize things inside lira as well:
* Moved test dir inside lira module, which simplifies importing of lira components inside tests and leaves less room for error
* Created requirements.txt, which Dockerfile uses to install Python dependencies. This makes it easier to run unit tests outside the docker container (saving the time of building the image), since you can build a virtual environment from the requirements.txt sufficient to run the tests.
* Moved code out of `lira/__init__.py` and into lira/lira.py, and simplified startup of lira
* Removed unused app.yaml, previously used for deploying lira to app engine
* Renamed green_box to lira throughout

PR #36, opened 2017-11-19
Read wdls from github and local disk

dshiga:
Enable Lira to read wdls from github and local disk and create dependencies zip on the fly.

I've kept the changes here small to reduce conflicts as much as possible with changes related to other PRs, like (#35) and https://github.com/broadinstitute/cromwell-tools/pull/2, which move cromwell_utils from lira to the cromwell-tools repo.

Once those other PRs are merged, I'll update this one to use cromwell-tools, to remove the read from bucket functionality, and to update the Lira config classes.


green_box/api/notifications.py
```diff
@@ -6,7 +6,7 @@
 from green_box_utils import gcs_utils
 from green_box_utils import cromwell_utils
 from green_box_utils import listener_utils
-
+import read_utils
```
ambrosejcarr: to me, "read" invokes a local operation. `io_utils`, maybe? I thought about `download_utils` but that has the opposite problem.

green_box/api/read_utils.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def make_zip(url_to_contents):
+    """
+    Given a map of urls and their contents, returns an in-memory zip file
+    containing each file. For each url, the part after the last slash is used
+    as the file name when writing to the zip archive.
+    """
+    buf = StringIO()
+    with zipfile.ZipFile(buf, 'w') as zip_buffer:
```
ambrosejcarr: Ah, clever. So, here the zipfile is being closed, but the StringIO() buffer stays open, so this produces a valid zipfile? I had a lot of trouble getting this to work because the Zipfile needs to close in order to write header records (python limitation).

green_box/api/read_utils.py
```diff
@@ -0,0 +1,76 @@
...
+def download_http(url):
+    """
+    Makes an http request for the contents at the given url and returns the response body.
+    """
+    response = requests.get(url)
+    response_str = response.text.encode('utf-8')
+    return response_str
+
+
+def read_local_file(path):
```
ambrosejcarr: I'm pretty sure that in the case of a local file you can just pass the open filehandle and avoid reading the whole file. 

I'm not sure if requests will then just read the whole file, or if it streams efficiently and doesn't necessarily bring the whole payload into memory at once.

ambrosejcarr (approved):
Minor comments. Looks good!

rexwangcc (approved):
Looks great!

green_box/api/read_utils.py
dshiga: Yeah, it seems that closing the ZipFile does not close the underlying StringIO object, so we accumulate data in the StringIO and then can pass it around to where it's needed.

green_box/api/read_utils.py
dshiga: Yes, you're right. I did it this way for consistency of reading via http and local files. In both cases, we read the full contents and store it in a map. These wdl files are tiny so it shouldn't be a performance issue. If it becomes one we can revisit it.

green_box/api/notifications.py
dshiga: Hmm, I don't have that association. I think of read primarily as being in opposition to write, regardless of location. In any case, this name is very temporary, since I want to move this code out of here and likely into cromwell-tools after #35 gets merged and there is a clear path to moving it. I'll keep it for now and it will change very soon.

ambrosejcarr: Fine with me! More just posted as a discussion point; sounds like changing would be moot based on what you're saying :)

PR #40, opened 2017-11-23
Bug fixes for cromwell inputs json

dshiga:
Fix two bugs in lira related to the inputs json file. These were introduced due to bad merge conflict resolution on my part when rebasing and merging #36.

### DIFF
--- a/lira/lira_utils.py
+++ b/lira/lira_utils.py
@@ -47,16 +47,20 @@ def extract_uuid_version_subscription_id(msg):
     return uuid, version, subscription_id
 
 
-def compose_inputs(workflow_name, uuid, version):
+def compose_inputs(workflow_name, uuid, version, env):
     """Create Cromwell inputs file containing bundle uuid and version.
 
     :param str workflow_name: The name of the workflow.
     :param str uuid: uuid of the bundle.
     :param str version: version of the bundle.
+    :param str env: runtime environment, such as 'dev', 'staging', 'int' or 'prod'.
     :return dict: A dictionary of workflow inputs.
     """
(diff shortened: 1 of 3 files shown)


lira/lira_utils.py
```diff
@@ -47,16 +47,19 @@ def extract_uuid_version_subscription_id(msg):
     return uuid, version, subscription_id
 
 
-def compose_inputs(workflow_name, uuid, version):
+def compose_inputs(workflow_name, uuid, version, env):
     """Create Cromwell inputs file containing bundle uuid and version.
 
     :param str workflow_name: The name of the workflow.
     :param str uuid: uuid of the bundle.
     :param str version: version of the bundle.
```
rexwangcc, line 55: `:param str env: runtime environment, such as 'dev', 'staging', 'int' or 'prod'.`

lira/lira_utils.py
```diff
@@ -47,16 +47,19 @@ def extract_uuid_version_subscription_id(msg):
...
     :return dict: A dictionary of workflow inputs.
     """
-    return {workflow_name + '.bundle_uuid': uuid,
-            workflow_name + '.bundle_version': str(version)}
+    return {
+        workflow_name + '.bundle_uuid': uuid,
+        workflow_name + '.bundle_version': str(version),
```
rexwangcc: If the version is `str` as the input parameter, maybe it is not necessary to `str()` it here? Although `str()` it would be safer there, it looks a little bit redundant. Probably this is my fault since I add the string conversion here in the first place.

rexwangcc (approved):
Minor comments, generally look good to me!

PR #41, opened 2017-11-23
Move code for making dependencies zip to cromwell-tools

dshiga:
See corresponding PR in cromwell-tools: https://github.com/broadinstitute/cromwell-tools/pull/3


lira/test/test_lira_utils.py
```diff
@@ -42,29 +42,6 @@ def test_compose_inputs(self):
         self.assertEqual(inputs['foo.bundle_version'], 'baz')
         self.assertEquals(inputs['foo.runtime_environment'], 'asdf')
 
-    def test_download_to_map(self):
-        """Test download_to_map with local files to ensure it builds the map of paths to contents correctly"""
-        urls = ['data/a.txt', 'data/b.txt']
-        urls_to_content = lira_utils.download_to_map(urls)
-        self.assertIn('data/a.txt', urls_to_content)
```
samanehsan, line 49: Can the files `a.txt` and `b.txt` be removed from lira?

samanehsan:
I have one question about the test files, otherwise this looks good!

dshiga:
Do you mean remove this test also, or test in a different way? The test
will break if we remove those two test data files.

lira/test/test_lira_utils.py
dshiga: oh right, yes no longer need them here

dshiga:
I removed those test files. Back to you @samanehsan!

samanehsan (approved):
Thanks for making that change! This looks good to me!

PR #46, opened 2017-12-19
Add travis yaml file for unit tests

dshiga:
This will make unit tests run in Travis on each push to GitHub. We already have unit tests running in Jenkins for Lira, however, I think it is preferable where possible to have our tests running in a publicly accessible place, and one that Mint team can manage ourselves. (Once this is merged we should turn off the Jenkins unit tests for Lira).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: python
+python:
+- '2.7'
+sudo: enabled
+services: docker
+script: cd lira/test && bash test.sh
+notifications:
+  slack:
+    on_success: change
+    on_failure: always
+    rooms:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,4 @@
+language: python
```
rexwangcc, line 1: I'd like to add:
```yaml
python:
  - 2.7
```
below to specify the python version, to be more accurate.

.travis.yml
```diff
@@ -0,0 +1,4 @@
...
+sudo: enabled
+services: docker
+script: cd lira/test && bash test.sh
```
rexwangcc: When we replace our current Jenkins with Travis, I think it would be great to enable notifications feature of Travis, say let it send notifications to slack channels:
```yaml
notifications:
  slack:
    on_success: change
    on_failure: always
    secure: xxx-xxx-xxx-xxx
```

rexwangcc (approved):
Minor comments, generally LGTM!

PR #48, opened 2018-01-04
Support analysis wdls with their own imports

dshiga:
Previously Lira was used only to launch monolithic analysis wdls. Now, we need to support launching analysis wdls that have their own imports.

To support this, the analysis_wdl field in the wdl config is now required to be a list (and is renamed to analysis_wdls). Also fixed an issue in `WdlConfig.__str__`, where the call to `.format` was expecting too many parameters.

See https://elastc.com/c/0v6eJH1B/465-updating-lira-to-handle-subworkflows

### DIFF
--- a/lira/config.py
+++ b/lira/config.py
@@ -3,7 +3,7 @@
 import logging
 
 
-class Config:
+class Config(object):
 
     def __init__(self, config_dictionary, flask_config_values=None):
         """abstract class that defines some useful configuration checks for the listener
@@ -17,7 +17,6 @@ class Config:
           those generated from a connexxion App
         """
         self.config_dictionary = config_dictionary
-        self._verify_fields()
 
         # make keys accessible in the namespace, grab any extra flask arguments
         for k, v in config_dictionary.items():
@@ -25,6 +24,8 @@ class Config:
         if isinstance(flask_config_values, dict):
             for k, v in flask_config_values.items():
                 setattr(self, k, v)
+
+        self._verify_fields()
 
     @property
     def required_fields(self):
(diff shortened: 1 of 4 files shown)


lira/config.py
```diff
@@ -3,7 +3,7 @@
 import logging
 
 
-class Config:
+class Config(object):
```
dshiga, line 6: This is necessary to allow calling super() in WdlConfig.

lira/config.py
```diff
@@ -26,6 +25,8 @@ def __init__(self, config_dictionary, flask_config_values=None):
             for k, v in flask_config_values.items():
                 setattr(self, k, v)
 
+        self._verify_fields()
```
dshiga, line 28: Now that `WdlConfig` is overriding `_verify_fields`, it is more convenient to have the `_verify_fields` call after attributes are set, since it means `WdlConfig.verify_fields` can refer to fields with the more convenient syntax, e.g. `self.analysis_wdls`.

lira/test/data/config.json
```diff
@@ -9,15 +9,22 @@
...
       "options_link": "https://raw.githubusercontent.com/HumanCellAtlas/pipeline-tools/master/adapter_pipelines/10x/options.json", 
       "wdl_default_inputs_link": "https://raw.githubusercontent.com/HumanCellAtlas/pipeline-tools/master/adapter_pipelines/10x/adapter_example_static.json", 
       "wdl_link": "https://raw.githubusercontent.com/HumanCellAtlas/pipeline-tools/master/adapter_pipelines/10x/adapter.wdl", 
       "workflow_name": "Adapter10xCount"
     }, 
     {
       "subscription_id": "dd17bb03-7634-47a4-9fd3-a55580ac3b1c", 
-      "analysis_wdl": "https://raw.githubusercontent.com/HumanCellAtlas/skylab/master/smartseq2_single_sample/ss2_single_sample.wdl", 
+      "analysis_wdls": [
+        "https://raw.githubusercontent.com/HumanCellAtlas/skylab/master/smartseq2_single_sample/ss2_single_sample.wdl",
```
rexwangcc: Should we use the master version of ss2 in the test? I feel like we need to revisit and change here to use a specific tagged version of ss2 later?

rexwangcc (approved):
A single comment, looks great to me!

PR #49, opened 2018-01-22
Use gunicorn

dshiga:
Use gunicorn instead of flask alone. See: https://elastc.com/c/rJJ6&YMM/358-use-gunicorn-or-other-server-instead-of-flask-alone

Also:
-Make port a parameter
-Improve some outdated scripts
-Use Ubuntu 16.04 instead of 17, since 16.04 has long term support

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,4 +22,4 @@ RUN pip install -r requirements.txt
 
 COPY . .
 
-CMD ["python", "lira/lira.py"]
+ENTRYPOINT ["bash", "start_lira.sh"]
--- a/README.md
+++ b/README.md
@@ -1,50 +1,94 @@
-# secondary-analysis
+# Lira
 
 | Slack Channel | Container | Linux-Python2.7 |
 |:-------:|:---------:|:---------:|
 | [![](https://img.shields.io/badge/slack-%23secondary--analysis-5BBC66.svg)](https://humancellatlas.slack.com/messages/secondary-analysis/) | [![Docker Repository on Quay](https://quay.io/repository/humancellatlas/secondary-analysis-lira/status "Docker Repository on Quay")](https://quay.io/repository/humancellatlas/secondary-analysis-lira) | [![Build Status](https://travis-ci.org/HumanCellAtlas/lira.svg?branch=master)](https://travis-ci.org/HumanCellAtlas/lira) |
 
-Prototype Secondary Analysis Service (Green Box) for the Human Cell Atlas. 
+Submits workflows to [Cromwell](https://github.com/broadinstitute/cromwell) in response to notifications.
 
-Whenever a bundle is uploaded to the storage service (Blue Box), if it matches a query registered with the listener it will send a notification to the listener's `/notifications` endpoint. 
-When the listener receives a notification, it processes data in that bundle using the workflow defined in `config.json` that has a matching `subscription_id`. 
-Currently, the listener launches either a smartseq2 or 10x count workflow to produce a bam and gene expression table. The analysis outputs are then uploaded to the storage service.
+Notifications contain a bundle id and version to identify a data bundle in the [Data Storage System](https://github.com/HumanCellAtlas/data-store). Notifications also contain a subscription id.
 
-[Design specs and prototypes for secondary analyses ("green box")](https://docs.google.com/document/d/1_VgySxINPbUsI0w-Gr4fV4DrHRSwdbCMf7b5sCB18uQ/edit?usp=sharing)
+When Lira receives a notification at its `/notifications` endpoint, it uses the subscription id to determine what type of workflow to submit, and uses the bundle id and version as inputs to that workflow.
 
-### Installation and Setup for Local Development
+[Design specs for the Secondary Analysis Service ("Green Box")](https://docs.google.com/document/d/1_VgySxINPbUsI0w-Gr4fV4DrHRSwdbCMf7b5sCB18uQ/edit?usp=sharing)
+
+## Building and running
+
+You can run Lira in docker or a Python virtual environment.
+
+### Using docker
 1. [Install Docker](https://docs.docker.com/engine/installation/#supported-platforms)
 2. Git clone this repository
-3. Add `config.json` (contains wdl configs and cromwell credentials) and `bucket-reader-key.json` (contains google cloud storage keys) to `/private/etc/secondary-analysis/`
-4. Build the docker container: `bash build_docker.sh dev test`
-5. Run the docker container: `bash run_docker.sh dev test`
+3. Create a `config.json` file (contains wdl configs and cromwell credentials). See an example at lira/test/data/config.json.
+4. Build the docker container: `bash build_docker.sh test`
+5. Run the docker container: `bash run_docker.sh test /path/to/config.json [PORT]`
 
-### Run unit tests
-There are two ways to do this, with and without Docker.
+### Virtual environment
+To run without docker, create a virtual environment.
 
-To run unit tests without building the docker image, you should create a virtual environment with the requirements for Lira:
+If you don't have pip installed, [install it first](https://pip.pypa.io/en/stable/installing/).
 
+Then install virtualenv with `pip install virtualenv`.
+
+Create a virtual environment for running Lira:
 ```
 virtualenv test-env
 source test-env/bin/activate
 pip install -r requirements.txt
 ```
 
-Then, from the root of the lira repo, do:
-```
-python -m unittest discover -v
-```
-This will run all of Lira's unit tests.
+There are two ways to run Lira in a virtual environment.
 
-You can also run the unit tests from inside the docker image, if you don't want to create a virtualenv.
-You can do this by running:
+#### gunicorn
+For production use, start Lira with:
+```
+bash start_lira.sh [PORT]
+```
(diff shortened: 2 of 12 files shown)


Dockerfile
```diff
@@ -20,4 +20,4 @@ RUN pip install -r requirements.txt
 
 COPY . .
 
-CMD ["python", "lira/lira.py"]
+ENTRYPOINT ["gunicorn", "lira.lira"]
```
rexwangcc: After changing from 17.04 to 16.04, it seems there is something weird about pip version. I'd add `RUN pip install --upgrade pip` before `RUN pip install wheel` if the pip version is 8 instead of 9. (See https://travis-ci.org/HumanCellAtlas/lira/builds/332047588 line 1227)

kubernetes/populate-listener-config-secret.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
-kubectl create secret generic listener-config \
---from-file=config=/etc/secondary-analysis/config.json \
---from-file=bucket-reader-key=/etc/secondary-analysis/bucket-reader-key.json
+config_name=$1
+config_file=$2
+
+kubectl create secret generic \
+    $config_name \
```
rexwangcc: I'm thinking about if we (not very possibly) want to deploy lira several times in a day, we may end up with having lots of config files such as `lira-config-2018-01-01-1`,`lira-config-2018-01-01-2`, ..., `lira-config-2018-01-01-n`, how about we use something like `kubectl create secret generic listener-config-$(date '+%Y-%m-%d-%H-%M-%S')` here?

lira/test/test.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-docker build -t gcr.io/broad-dsde-mint-dev/lira:test ../..
+docker build -t lira:test ../..
```
rexwangcc: Love this change, eliminated a lot of possible confusion, I will change this line in my PR accordingly.

README.md
```diff
@@ -12,38 +12,82 @@ Currently, the listener launches either a smartseq2 or 10x count workflow to pro
 
 [Design specs and prototypes for secondary analyses ("green box")](https://docs.google.com/document/d/1_VgySxINPbUsI0w-Gr4fV4DrHRSwdbCMf7b5sCB18uQ/edit?usp=sharing)
 
-### Installation and Setup for Local Development
+## Building and running
```
rexwangcc: The main title of this Readme is still "secondary-analysis", as well as the summary under the title, could we also change it to Lira and replace the general description of Green-Box to a description of Lira?

rexwangcc: Never mind, I found there is another ticket about updating readmes for all repos once and for all, don't need to do it on this ticket.

kubernetes/listener-deployment.yaml
```diff
@@ -24,9 +24,7 @@ spec:
       volumes:
       - name: listener-config
         secret:
-          secretName: listener-config
+          secretName: listener-config-2018-01-22
```
rexwangcc, line 27: Not sure if Year-Month-Date is atomic enough, maybe add Hour-Minute-Second as well? I dropped another comment about this below.

kubernetes/populate-listener-config-secret.sh
rexwangcc: Update: Just came up with a better way:
```
kubectl create secret generic ${config_name:-$(date '+%Y-%m-%d-%H-%M-%S')} --from-file=config=$config_file
```
By using the default variable value here, we have the flexibility to pass a name to the config file and at the same time, give the date-hour style suffix to it to provide enough atomicity. What do you think?

kubernetes/populate-listener-config-secret.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
-kubectl create secret generic listener-config \
---from-file=config=/etc/secondary-analysis/config.json \
---from-file=bucket-reader-key=/etc/secondary-analysis/bucket-reader-key.json
+config_name=$1
+config_file=$2
```
rexwangcc: Do we also want to add error handling for the `config_file` parameter?

lira/lira.py
```diff
@@ -8,30 +8,29 @@
...
-                                          scopes=['https://www.googleapis.com/auth/devstorage.read_only'])
-    app = connexion.App(__name__)
+app = connexion.App(__name__)
 
-    config_path = os.environ['listener_config']
-    with open(config_path) as f:
-        app.app.config = config.ListenerConfig(json.load(f), app.app.config)
-        app.app.gcs_client = gcs_client
+# Defining 'application' simplifies calling it from gunicorn
+application = app.app
```
rexwangcc: I might miss something, but I can't see this `application` variable been used anywhere?

run_docker.sh
```diff
@@ -1,4 +1,33 @@
...
+if [ $error==1 ]; then
+    echo -e "\nUsage: bash run_docker.sh TAG /path/to/config.json [PORT]\n"
+    exit 1
+fi
+
+mounted_config=/etc/secondary-analysis/config.json
+
+# Remove any previously started containers
+docker stop lira
+docker rm lira
```
rexwangcc: If there are no previously started containers, this will output an error msg to the console, is this what we expected?

rexwangcc (commented):
Dropped a few comments, generally looks good!

lira/lira.py
dshiga: Not used here, but when gunicorn runs Lira (or any wsgi application) it looks for a variable called "application".  I'll clarify the comment.

kubernetes/populate-listener-config-secret.sh
dshiga: Good point, added

kubernetes/populate-listener-config-secret.sh
dshiga: Okay, may be useful to have hour and minute. However, the name is getting very long and I don't think we need seconds. If we are configuring Lira more than once per minute, I think we're doing something wrong. I think we can just hard code the prefix of the name to be listener-config for now, as you did in your PR.

run_docker.sh
dshiga: I think that's okay, will add an echo line to clarify

kubernetes/populate-listener-config-secret.sh
rexwangcc: I agree with that we don't need seconds here. And since my PR was already merged in, would you mind changing the `kubernetes/populate-listener-config-secret.sh`, removing the seconds I used there? Thanks!

lira/lira.py
rexwangcc: Thanks for the explanation!!

README.md
```diff
@@ -1,50 +1,94 @@
-# secondary-analysis
+# Lira
```
dshiga, line 1: Modified this intro section and replaced "listener" with "Lira" throughout README.

lira/lira.py
```diff
@@ -8,30 +8,29 @@
...
-    gcs_client = gcs_utils.GoogleCloudStorageClient(key_location='/etc/secondary-analysis/bucket-reader-key.json',
-                                          scopes=['https://www.googleapis.com/auth/devstorage.read_only'])
-    app = connexion.App(__name__)
+app = connexion.App(__name__)
 
-    config_path = os.environ['listener_config']
-    with open(config_path) as f:
-        app.app.config = config.ListenerConfig(json.load(f), app.app.config)
-        app.app.gcs_client = gcs_client
+# 'application' is not used in this file, but is used by gunicorn
```
dshiga, line 25: Clarified this comment

kubernetes/populate-listener-config-secret.sh
```diff
@@ -1,5 +1,13 @@
 #!/usr/bin/env bash
 
-kubectl create secret generic listener-config-$(date '+%Y-%m-%d-%H-%M-%S') \
---from-file=config=/etc/secondary-analysis/config.json \
---from-file=bucket-reader-key=/etc/secondary-analysis/bucket-reader-key.json
+config_file=$1
```
dshiga, line 3: Added error handling, removed name parameter, added timestamp to name.

lira/test/test.sh
```diff
@@ -3,4 +3,4 @@
 docker build -t lira:test ../..
 
 # Run unit tests in docker container
-docker run -e listener_config=config.json lira:test bash -c "python -m unittest discover -v"
+docker run -e listener_config=config.json --entrypoint python lira:test -m unittest discover -v
```
dshiga, line 6: Removed `"usr/bin/python"` and simplified to python

dshiga:
Hi @rexwangcc, I made a few changes to the README, to populate-listener-config-secret.sh and lira.py. Unfortunately, I wasn't able to separate those out into a new commit because there were conflicts in the earlier commits when I rebased on master. But I commented on the lines of code that changed.

kubernetes/populate-listener-config-secret.sh
rexwangcc: Thanks for this change!

rexwangcc (approved):
Looks great to me! Thanks!

dshiga:
jenkins retest

PR #51, opened 2018-01-25
Streamline setting up a new lira

samanehsan:
Currently in order to set up a new lira, you need to go through 15+ different steps: https://github.com/HumanCellAtlas/lira/tree/master/kubernetes#create-a-new-kubernetes-listener-deployment

These scripts reduce the number of steps it takes to set up a new lira to: 
1. Run `listener-deployment-setup.sh` to create a new gcloud service account, make subscriptions to the data storage service and generate a config.json file for lira
2. Run `kubernetes.sh` to create a deployment, service and ingress for the new lira and set up a DNS record to point to the static IP address


dshiga:
Thanks for putting this up early!

samanehsan:
@dshiga, the `listener-deployment-setup.sh` and `listener-config.json.ctmpl` files are ready for review

kubernetes/populate-listener-config-secret.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+cert_dir=$1
+
 kubectl create secret generic listener-config-$(date '+%Y-%m-%d-%H-%M-%S') \
---from-file=config=/etc/secondary-analysis/config.json \
---from-file=bucket-reader-key=/etc/secondary-analysis/bucket-reader-key.json
+--from-file=config=$cert_dir/config.json \
+--from-file=bucket-reader-key=$cert_dir/bucket-reader-key.json
```
dshiga: Lira no longer needs the bucket reader key, so this line can be removed.

kubernetes/populate-listener-config-secret.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+cert_dir=$1
+
 kubectl create secret generic listener-config-$(date '+%Y-%m-%d-%H-%M-%S') \
```
dshiga: I would suggest we not use seconds in the timestamp, seems excessive and cutting it out will make it a little more readable.

kubernetes/listener-deployment-setup.sh
```diff
@@ -0,0 +1,59 @@
...
+gcloud config set project ${gcloud_project}
+
+# Create service account & key
+echo "Creating bluebox-subscription-manager service account"
+gcloud iam service-accounts create bluebox-subscription-manager --display-name=bluebox-subscription-manager
+gcloud iam service-accounts keys create ${bluebox_subscription_key_path} --iam-account=bluebox-subscription-manager@${gcloud_project}.iam.gserviceaccount.com
+
+# Add key to vault
+key_file_path=$(echo "${bluebox_subscription_key_path/#$HOME//root}")
+docker run -it --rm -v $HOME:/root broadinstitute/dsde-toolbox vault write secret/dsde/mint/${env}/listener/bluebox-subscription-manager-${env}-key.json @${key_file_path}
```
dshiga: Might be able to simplify here if instead of mounting `$HOME`, we mount `$bluebox_subscription_key_path` (or make it `bluebox_subscription_key_dir` above and mount that) -- I think you could then avoid the `#$HOME//root` magic and this would make it a little more generic in that people could choose to use another directory besides `$HOME` here.

kubernetes/listener-deployment-setup.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+
+#Create subscriptions to the HCA Data Storage Service for the 10x and ss2 pipelines and generate a config.json file
+#for a new lira deployment that will run the 10x and ss2 pipelines.
+
+env=$1  #name of the deployment environment (e.g. staging)
```
dshiga: Might be helpful to check for missing parameters here, similar to this: https://github.com/HumanCellAtlas/lira/blob/master/run_docker.sh#L6

kubernetes/listener-deployment-setup.sh
```diff
@@ -0,0 +1,59 @@
...
+gcloud iam service-accounts keys create ${bluebox_subscription_key_path} --iam-account=bluebox-subscription-manager@${gcloud_project}.iam.gserviceaccount.com
+
+# Add key to vault
+key_file_path=$(echo "${bluebox_subscription_key_path/#$HOME//root}")
+docker run -it --rm -v $HOME:/root broadinstitute/dsde-toolbox vault write secret/dsde/mint/${env}/listener/bluebox-subscription-manager-${env}-key.json @${key_file_path}
+
+# Create subscriptions for ss2 and 10x queries
+green_url=https://pipelines.${env}.data.humancellatlas.org/notifications
+key_file=${bluebox_subscription_key_path}
+listener_secret=$(docker run -it --rm -v $HOME:/root broadinstitute/dsde-toolbox vault read -field=notification_token secret/dsde/mint/${env}/listener/listener_secret)
```
dshiga: We could make this a bit more flexible by removing the volume mount and instead adding `-e VAULT_TOKEN=${vault_token}`. (The volume mount is only used here to read the token from the local file system.) That way people can store the token wherever they choose, not necessarily in `$HOME`.

Alternatively, you could mount just the token file like you do in the other script:
`-v ${VAULT_TOKEN_FILE}:/root/.vault-token`.

We probably should decide whether we want the script to take:
A) a path to the token file or
B) the token contents
and use just one or the other everywhere. Right now both are used in different places.

dshiga (commented):
This looks really good! I suggested a few changes in comments.

samanehsan:
Thanks for your comments @dshiga! The rest of the code is ready to review now too!

samanehsan:
jenkins retest

kubernetes/listener-deployment.sh
```diff
@@ -0,0 +1,103 @@
...
+printf "\ncreating static IP address"
+gcloud compute addresses create listener --global
+
+# Create DNS zone
+zone=mint-${env}
+dns_name=mint-${env}.broadinstitute.org.
+gcloud dns managed-zones create ${zone} --dns-name ${dns_name} --description "mint-${env}.broadinstitute.org zone"
+
+# Add a record-set that points to the static ip address
+ip_address=echo gcloud compute addresses describe listener --global --format='value(address)'
```
dshiga: When I tried running this command, it printed the address but didn't save it in ip_address. I think you need something like:
```
ip_address=$(gcloud compute addresses describe listener --global --format='value(address)')
echo $ip_address
```

dshiga (commented):
This looks great!

I think we will want to make listener-deployment.sh more generic over time.

The way it is written works for us, but has some hard-coded values that won't work for others outside of mint, broad, or hca.

To make it more generic, I think we'll eventually want more input parameters (or perhaps an input json file) instead of those hard-coded values.

For now, it might be worth seeing if there are uses of "mint", "broad", or "hca" that could be replaced with variables that are defined once at the top of the script, at least as long as this doesn't hurt the clarity of the script.

Then in a later ticket we could try to make the script fully generic.

Also, maybe think furtehr about organization within this and the other scripts. Which commands will always make sense to run together, and which are more separable? Are there groups of commands that you could imagine people might want to skip in certain scenarios, whether us or others outside Mint, Broad, or HCA? Thinking through the grouping now will help set the stage in future if we want to add conditionals or break things out into separate scripts.

kubernetes/listener-deployment.sh
```diff
@@ -0,0 +1,103 @@
...
+env=$1
+get_certs=$2
+vault_token=$3
+cert_dir=$4
+fullchain_filename=$5
+privkey_filename=$6
+docker_tag=$7 #lira docker version tag
+
+# vault token file
+VAULT_TOKEN_FILE=${VAULT_TOKEN_FILE:-"$HOME/.vault-token"}
```
dshiga: Use of all caps vs lowercase for variable names feels a bit arbitrary -- I'm not hugely bothered if there is a mixture but it's probably best to be more consistent one way or the other, with fewer exceptions unless there is a compelling reason for the difference.

kubernetes/listener-deployment.sh
```diff
@@ -0,0 +1,103 @@
...
+        -v $HOME:/root:rw \
+        broadinstitute/dsde-toolbox vault auth -method=github token=$vault_token
+
+    export AWS_ACCESS_KEY_ID=$(docker run -it --rm -v $HOME:/root broadinstitute/dsde-toolbox vault read -field=AWS_ACCESS_KEY_ID secret/dsde/mint/${env}/listener/aws-access-key)
+
+    export AWS_SECRET_ACCESS_KEY=$(docker run -it --rm -v $HOME:/root broadinstitute/dsde-toolbox vault read -field=AWS_SECRET_ACCESS_KEY secret/dsde/mint/${env}/listener/aws-access-key)
+
+    # Get existing certificates from vault
+    printf "\nget certificates from vault"
+    docker run -it --rm -v $HOME:/root broadinstitute/dsde-toolbox vault read -format=json -field=value secret/dsde/mint/${env}/listener/$fullchain_filename > $cert_dir/fullchain.pem
```
dshiga: As mentioned in a comment on listener-deployment-setup.sh, I would avoid requiring mounting of the `$HOME` directory and instead pass in the token to the docker image. It's one less requirement that people will have to conform to.

kubernetes/listener-deployment.sh
samanehsan: That's a good point! I'll switch to using all caps since that's what the render-ctmpl.sh script is using.

kubernetes/listener-deployment-setup.sh
```diff
@@ -0,0 +1,81 @@
...
+elif [ -z $PIPELINE_TOOLS_PREFIX ]; then
+    echo -e "\nYou must specify the pipeline-tools prefix"
+    error=1
+elif [ -z $SS2_PREFIX ]; then
+    echo -e "\nYou must specify the SmartSeq2 prefix"
+    error=1
+elif [ -z $TENX_PREFIX ]; then
+    echo -e "\nYou must specify the 10x prefix"
+    error=1
+fi
```
dshiga, line 39: You should exit the script here if parameters are missing.

dshiga (approved):
Just need to add logic to exit the scripts early when you detect parameters missing. Once that's done I think this is ready to merge.

kubernetes/README.md
```diff
@@ -2,29 +2,17 @@
...
-11. Create deployment: `kubectl create -f listener-deployment.yaml --record --save-config`
-12. Create service: `kubectl create -f listener-service.yaml --record --save-config`
-13. Create ingress: `kubectl create -f listener-ingress.yaml --record --save-config`
-14. Create static ip address: `gcloud compute addresses create listener --global`
-15. In the Google Cloud Platform console, go to the Network Services/Cloud DNS page and add an "A" record named "listener101.mint-[env].broadinstitute.org" that points to the static ip address
-16. In AWS, go to route53, click on hosted zones, then data.humancellatlas.org. Add or modify a CNAME record named
-    pipelines.[env].data.humancellatlas.org and set the value to listener101.mint-[env].broadinstitute.org.
-    For prod this should be pipelines.data.humancellatlas.org.
+1. Create a new gcloud project for the deployment
+2. Run the `listener-deployment-setup.sh` script to create subscriptions to the HCA data storage service and
```
dshiga, line 11: Would be good to give some indication that the setup script assumes you're setting it up for mint/broad/hca, not expected to work otherwise. Maybe also indicate that at the top of the script too.

dshiga:
👍

PR #53, opened 2018-02-02
Cache wdls

dshiga:
Cache wdls and static inputs in Lira instead of fetching them from GitHub on each workflow launch. See:
https://elastc.com/c/zCXY1Hjc/538-cache-wdls-and-static-inputs-in-lira

Uses functools.lru_cache library, which is only available in Python 3. (I fixed the remaining compatibility issues in Python 3, so we can run Lira in Python 3 now.) In Python 2, caching is disabled, using a no-op decorator instead of the lru_cache one.

Also renamed listener to Lira throughout.

### DIFF
--- a/lira/api/notifications.py
+++ b/lira/api/notifications.py
@@ -6,13 +6,15 @@ from flask import current_app
 from cromwell_tools import cromwell_tools
 from lira import lira_utils
 
+logger = logging.getLogger("Lira | {module_path}".format(module_path=__name__))
+
 
 def post(body):
-    # Check authentication
-    logger = logging.getLogger("Lira | {module_path}".format(module_path=__name__))
-    green_config = current_app.config
+    """Process notification and launch workflow in Cromwell"""
 
-    if not lira_utils.is_authenticated(connexion.request.args, green_config.notification_token):
+    # Check authentication
+    lira_config = current_app.config
+    if not lira_utils.is_authenticated(connexion.request.args, lira_config.notification_token):
         time.sleep(1)
         return lira_utils.response_with_server_header(dict(error='Unauthorized'), 401)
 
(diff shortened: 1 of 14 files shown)


lira/api/notifications.py
```diff
@@ -23,43 +29,83 @@ def post(body):
...
 
     # Prepare inputs
-    inputs = lira_utils.compose_inputs(wdl.workflow_name, uuid, version, green_config.env)
+    inputs = lira_utils.compose_inputs(wdl_config.workflow_name, uuid, version, lira_config.env)
     cromwell_inputs_file = json.dumps(inputs)
 
+    cromwell_submission = prepare_submission(wdl_config, lira_config.submit_wdl)
+    logger.info(prepare_submission.cache_info())
+
+    dry_run = getattr(lira_config, 'dry_run', False)
```
rexwangcc: This "dry_run" mode is designed for only testing purposes?

lira/api/notifications.py
```diff
@@ -23,43 +29,83 @@ def post(body):
...
+
+
+@lru_cache(maxsize=None)
+# Setting maxsize to None here means each unique call to prepare_submission (defined by parameters used)
+# will be added to the cache without evicting any other items. Additional non-unique calls
+# (parameters identical to a previous call) will read from the cache instead of actually
+# calling this function. This means that the function will be called no more than once for
+# each wdl config, but we can have arbitrarily many wdl configs without running out of space
+# in the cache.
+def prepare_submission(wdl_config, submit_wdl):
```
rexwangcc: I'm wondering if we should add another endpoint to Lira that implements cache_clear() function so we could have some control of our cache. But that should be another ticket.

rexwangcc: Besides, maybe we should also wrap the lru_cache to add limits to the cache pool such as `use_mem_up_to`? Since we are setting `maxsize=None`, the cache can grow without bound now, and LRU is only memory aware when least recent used calls are removed from the pool.

Edit: Nevermind, I just noticed that this cache pool would never grow too large.

lira/lira_config.py
```diff
@@ -53,21 +54,16 @@ def __eq__(self, other):
...
         result = ''
         for v in self.required_fields:
             field_value = getattr(self, v)
-            if isinstance(field_value, (str, unicode, int, list)):
-                result += str(field_value)
-            elif isinstance(field_value, Config):
-                result += field_value.to_string()
-            else:
-                raise ValueError('Config fields must either be strings or config objects')
+            result += str(field_value)
```
rexwangcc, line 62: For completeness, it would be great to also implement the `__repr__(self)` function with the same code. (optional)

lira/api/notifications.py
dshiga: Good thoughts. Keep in mind that all of the files that Lira caches are from versioned releases in GitHub, and reading from the same url should always return the exact same file. So clearing and refilling the cache would just put the same files in it again. If we want to use different files, it requires pointing to different urls, which requires restarting Lira anyway to use a different configuration.

Although theoretically the cache can grow without bound, in practice it is limited by the number of different wdl configs in Lira's configuration. We will get at most one cache entry for each wdl config (unit tests are testing that). Each cache entry should be at most a few KB, since it's just some wdl and input json files. So even if we had 100 different pipelines, the cache would take up < 1 MB.

rexwangcc: Thanks for the explanation here, I realized I miss understood the cache here. Yeah, practically the cache pool here will never grow too large.

rexwangcc (approved):
Looks great to me. Dropped some comments, but generally looks amazing! Please fix the tests and conflicts before merge in.

lira/api/notifications.py
dshiga: Yes, I debated whether to keep this. My reasons in favor of keeping are:
-It will greatly facilitate testing, especially for things like caching and load testing, where we might want to hammer Lira with thousands of messages in a short time, without being forced to launch all those workflows in Cromwell.
-Lots of command line tools have a "dry run" mode, and this is not so different.
-We may well need other modes in future, for example, a mode where Lira writes to a queue instead of submitting directly to Cromwell. In this case, "dry run" mode might be just one of several modes we support and having the basic code structure in place to choose between different modes will then be more generally useful.
-I've added warnings to the logs to reduce the likelihood that someone could run in this mode and not realize it.

rexwangcc: Thanks for the explanation, I personally voted for keeping this mode. And we could always refactor the code later if we want to add support for more modes to Lira.

lira/api/notifications.py
samanehsan: Some files in the lira config are pointing to the master branch instead of a versioned release (e.g. submit.wdl: `https://raw.githubusercontent.com/HumanCellAtlas/pipeline-tools/master/adapter_pipelines/submit.wdl`), so it will return the cached version even if we push a change to the master branch. Maybe the solution here is to point to a specific tag in pipeline-tools instead.

samanehsan:
This looks good to me @dshiga! I just have a comment about updating the lira config so that all of the file paths point to tagged versions.

lira/api/notifications.py
dshiga: Yes, I definitely think we should make all the configs point to tagged releases. Now that you mention this, I have found the ability to run without caching to be useful in the past for testing, so I've tweaked this PR so that caching can be configurably turned off (it's on by default).

lira/api/notifications.py
```diff
@@ -23,43 +30,96 @@ def post(body):
...
+    """Calls prepare_submission. In Python 3 these calls are cached.
+    In Python 2, the calls are not cached, since the functools.lru_cache decorator is not
+    available (a no-op version of the decorator is used instead)."""
+    return prepare_submission(wdl_config, submit_wdl)
+
+
+# Having an undecorated prepare_submission function is needed to support
+# the ability to turn caching on or off via Lira's configuration.
+def prepare_submission(wdl_config, submit_wdl):
+    """Load into memory all static data needed for submitting a workflow to Cromwell"""
```
samanehsan: There are two situations where we do not want to use `lru_cache` that are being handled differently (the noop decorator when using python 2, and calling a non-decorated function when `lira_config.cache_wdls = False`. I'm wondering if there's a way to consolidate this.

What do you think about adding a decorator that handles both situations by conditionally applying the `lru_cache` decorator if `lira_config.cache_wdls = True` and `lru_cache is not None` (if it's set to `None` on `ImportError`)?

samanehsan:
Thanks @dshiga! I just have one comment about possibly consolidating some of the logic around turning off caching.

samanehsan (approved):
I think this is good to merge once the conflicts are resolved and the integration tests are passing!

PR #56, opened 2018-02-09
Register service account for CaaS

samanehsan:
These scripts set up the required user accounts for lira to talk to cromwell-as-a-service. It will: 
- Create a service account and key for a given gcloud project
- Add the key to vault
- Register a service account user in firecloud
- Register a service account user in SAM
- Create a workflow-collection resource in SAM
- Create a group in firecloud to manage permissions
- Add reader and writer policies to the collection, and add the firecloud user group to the writer policy

### DIFF
--- a/kubernetes/caas_login.sh
+++ b/kubernetes/caas_login.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+# This script creates a service account in the specified gcloud project and registers it in FireCloud and SAM for use with Cromwell-as-a-service (CaaS).
+# Note: CaaS is only available in dev, so this script uses the develop versions of FireCloud and SAM
+
+GCLOUD_PROJECT=$1  # The gcloud project in which to create the service account
+ENV=$2  # The deployment environment associated with the gcloud project
+KEY_DIR=$3  # Where to save the service account key
+WORKFLOW_COLLECTION_ID=$4  # The name of the workflow-collection to create in SAM (e.g. "dev-workflows")
+FIRECLOUD_CONTACT_EMAIL=$5  # A contact email to use for the FireCloud service account
+FIRECLOUD_GROUP_NAME=$6  # The name of the user group to create in FireCloud to control workflow collection permissions (e.g. "write-access")
+VAULT_TOKEN_FILE=${VAULT_TOKEN_FILE:-"$HOME/.vault-token"}
+FIRECLOUD_URL="https://firecloud.dsde-dev.broadinstitute.org"
+
+
+#Set gcloud project
+gcloud config set project ${GCLOUD_PROJECT}
+
+echo "Creating CaaS service account and key"
+KEY_FILE_PATH=${KEY_DIR}/caas-${ENV}-key.json
+IAM_ACCOUNT_EMAIL=caas-account@${GCLOUD_PROJECT}.iam.gserviceaccount.com
+gcloud iam service-accounts create caas-account --display-name=caas-account
+gcloud iam service-accounts keys create ${KEY_FILE_PATH} --iam-account=${IAM_ACCOUNT_EMAIL}
+
+# Add key to vault
+docker run -it --rm -v ${VAULT_TOKEN_FILE}:/root/.vault-token -v ${KEY_DIR}:/keys broadinstitute/dsde-toolbox vault write secret/dsde/mint/${ENV}/listener/caas-${ENV}-key.json @/keys/caas-${ENV}-key.json
+
+# Clone firecloud-tools repo
+git clone git@github.com:broadinstitute/firecloud-tools.git
+cd firecloud-tools
+
+echo "Register the service account for use in Firecloud"
+./run.sh scripts/register_service_account/register_service_account.py -j ${KEY_FILE_PATH} -e ${FIRECLOUD_CONTACT_EMAIL} -u ${FIRECLOUD_URL}
+
+cd ..
+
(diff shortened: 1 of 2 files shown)


kubernetes/caas_login.sh
```diff
@@ -0,0 +1,37 @@
...
+# Note: CaaS is only available in dev, so this script uses the develop versions of FireCloud and SAM
+
+GCLOUD_PROJECT=$1  # The gcloud project in which to create the service account
+ENV=$2  # The deployment environment associated with the gcloud project
+KEY_DIR=$3  # Where to save the service account key
+WORKFLOW_COLLECTION_ID=$4  # The name of the workflow-collection to create in SAM (e.g. "dev-workflows")
+FIRECLOUD_CONTACT_EMAIL=$5  # A contact email to use for the FireCloud service account
+FIRECLOUD_GROUP_NAME=$6  # The name of the user group to create in FireCloud to control workflow collection permissions (e.g. "write-access")
+VAULT_TOKEN_FILE=${VAULT_TOKEN_FILE:-"$HOME/.vault-token"}
+FIRECLOUD_URL="https://firecloud.dsde-dev.broadinstitute.org"
```
rexwangcc: Is it necessary to parameterize this url as well? e.g. `"https://firecloud.dsde-${ENV}.broadinstitute.org"`

kubernetes/sam_registration.py
```diff
@@ -0,0 +1,65 @@
...
+import argparse
+import requests
+import logging
+from firecloud import api as firecloud_api
+from oauth2client.service_account import ServiceAccountCredentials
+
+
+def register_service_account(json_credentials, workflow_collection_id):
+    # Get bearer token from service account key
+    # From https://github.com/broadinstitute/firecloud-tools/blob/master/scripts/register_service_account/register_service_account.py
```
rexwangcc: It would be great if you could include the effective period/expiration of the Bearer token in the comment.

kubernetes/sam_registration.py
```diff
@@ -0,0 +1,65 @@
...
+    scopes = ['https://www.googleapis.com/auth/userinfo.profile', 'https://www.googleapis.com/auth/userinfo.email']
+    credentials = ServiceAccountCredentials.from_json_keyfile_name(json_credentials, scopes=scopes)
+    headers = {"Authorization": "bearer " + credentials.get_access_token().access_token}
+    headers["User-Agent"] = firecloud_api.FISS_USER_AGENT
+
+    logging.info('Register service account in SAM')
+    sam_url = "https://sam.dsde-dev.broadinstitute.org"
```
rexwangcc: Is it necessary to make this as a parameter of the function?

kubernetes/sam_registration.py
```diff
@@ -0,0 +1,65 @@
...
+
+def register_service_account(json_credentials, workflow_collection_id):
+    # Get bearer token from service account key
+    # From https://github.com/broadinstitute/firecloud-tools/blob/master/scripts/register_service_account/register_service_account.py
+    scopes = ['https://www.googleapis.com/auth/userinfo.profile', 'https://www.googleapis.com/auth/userinfo.email']
+    credentials = ServiceAccountCredentials.from_json_keyfile_name(json_credentials, scopes=scopes)
+    headers = {"Authorization": "bearer " + credentials.get_access_token().access_token}
+    headers["User-Agent"] = firecloud_api.FISS_USER_AGENT
+
+    logging.info('Register service account in SAM')
```
rexwangcc, line 17: By default the logging level is set to `WARNING` so all of the information here cannot be seen from the script. Maybe it is better  to get a logger and set the logging level lower to `INFO`, so the messages could be visible from the bash script?

kubernetes/sam_registration.py
```diff
@@ -0,0 +1,65 @@
...
+    sam_url = "https://sam.dsde-dev.broadinstitute.org"
+    requests.post(sam_url + "/register/user", headers=headers)
+
+    logging.info('Create new workflow-collection')
+    workflow_collection_url = "{}/api/resource/workflow-collection/{}".format(sam_url, workflow_collection_id)
+    requests.post(workflow_collection_url, headers=headers)
+
+    logging.info('Create group in firecloud')
+    firecloud_groups_url = "https://firecloud-orchestration.dsde-dev.broadinstitute.org/api/groups/{}".format(args.firecloud_group_name)
```
rexwangcc: Another place about parameterizing?

rexwangcc (commented):
Thanks for doing this, this is really great! @samanehsan I dropped a few comments.

kubernetes/caas_login.sh
samanehsan: Right now caas is only available in dev so firecloud and sam also need to use dev. But now that I think of it, if that changes in the future to work in production it's better not to hard-code this!

kubernetes/sam_registration.py
samanehsan: I set the logging level in the script here: https://github.com/HumanCellAtlas/lira/blob/dae3c908c6cb7e7ec3858dafab458f94bdf51094/kubernetes/sam_registration.py#L59

rexwangcc: Oh, I see! Thanks!

rexwangcc (approved):
LGTM!

samanehsan:
Thanks for taking a look @rexwangcc!

PR #57, opened 2018-02-15
Use DCP central logging

dshiga:
Adds script to set up DCP central logging and invokes it from main setup script.

Also, modify main setup script so that errors are printed for all missing parameters, not just the first missing one.

Suppress logs that print request query params.

See:
https://elastc.com/c/A9Q*9L4w/430-link-into-dcp-wide-logging-system

### DIFF
--- a/kubernetes/create-logging-sink.sh
+++ b/kubernetes/create-logging-sink.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+# Creates a sink for exporting logs from Stackdriver.
+# Lira logs get sent to Stackdriver when running in Google Cloud Platform.
+# Creating the sink allows the logs to be exported automatically to a
+# Google Pub/Sub topic that feeds them into the DCP's central logging infrastructure.
+
+# Parameters
+
+# sink_name
+# An arbitrary identifier that serves to uniquely identify
+# a logging export sink within a Google Cloud Project (there could be several in a project).
+
+# destination
+# This tells Google where to send the logs.
+# Several kinds of destinations are supported by Google, but for integration with
+# DCP centralized logging, this must point to the Pub/Sub topic that the central
+# logging system is monitoring, of the form
+# pubsub.googleapis.com/projects/<logs_project_id>/topics/<topic_name>
+# logs_project_id and topic_name can be obtained from the DCP logging system administrator.
+
+sink_name=$1
+destination=$2
+
+error=0
+if [ -z $sink_name ]; then
+  printf "\nYou must provide a sink name\n"
+  error=1
+fi
+if [ -z $destination ]; then
+  printf "\nYou must provide a destination\n"
+  error=1
+fi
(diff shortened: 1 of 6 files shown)


kubernetes/create-logging-sink.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+sink_name=$1
```
samanehsan: Can you add a brief comment at the top of the script to describe what it's doing? Also some comments about what the parameters are would be helpful as well

lira/lira.py
```diff
@@ -26,6 +26,10 @@
 # 'application' is not used in this file, but is used by gunicorn
 application = app.app
 
+# Turn off logs that print query params, since we want to exclude those from the logs
+logging.getLogger('werkzeug').setLevel(logging.ERROR)
+logging.getLogger('connexion.decorators.validation').setLevel(logging.INFO)
```
samanehsan: Can you clarify how this excludes logs that contain query params? Do those logs have a logging.DEBUG level?

lira/lira.py
```diff
@@ -26,6 +26,10 @@
 # 'application' is not used in this file, but is used by gunicorn
 application = app.app
 
+# Turn off logs that print query params, since we want to exclude those from the logs
+logging.getLogger('werkzeug').setLevel(logging.ERROR)
```
samanehsan: Would we exclude any useful/important logging info by raising the level higher? Is it possible to use a custom formatter here like you mentioned on the ticket?

lira/lira.py
dshiga: Done.

lira/lira.py
dshiga: Good point. I have modified things so that we can configurably set these log levels. They default to ERROR and INFO respectively, but adding entries to Lira's config json can boost those back to DEBUG or whatever the user wants.

This should be a temporary solution, until we start using a different method for authenticating notifications.

lira/lira_config.py
```diff
@@ -103,9 +103,16 @@ class LiraConfig(Config):
...
 
-        # Default value that can be overridden
+        # Setting default values that can be overridden
         self.cache_wdls = True
 
+        # Setting the following log levels prevents log messages that
+        # print query params in the log.
+        # Werkzeug has INFO messages that log the url including query params of each request.
+        # The Connexion validator has DEBUG messages that do the same.
+        self.log_level_werkzeug = 'ERROR'
```
samanehsan: Maybe the default here should be `WARNING` (if it doesn't expose query params), since that's one level above `INFO`: https://docs.python.org/3.6/library/logging.html#logging-levels

samanehsan (approved):
I just have one suggestion about changing the default logging level, otherwise I think this is good to merge after the tests pass!

dshiga:
jenkins retest

PR #58, opened 2018-02-20
Fix excessive logging

dshiga:
Kubernetes does health checks at the root url "/" by default, and it was doing it many times per second, creating a lot of noise in the logs.

This PR modifies the deployment ctmpl file to explicitly define a health check ("readinessProbe") at /health instead, at intervals of 10 seconds. This will greatly reduce the noise in the logs.

It should also fix a problem with logging that was causing all logs in Stackdriver to be marked as "error" level. (This was due to all logs being written to stderr, caused by setting the log level to "DEBUG".) We are still logging at DEBUG level here, but some added code causes anything at INFO and below to go to stdout instead of stderr.

See:
https://elastc.com/c/A9Q*9L4w/430-link-into-dcp-wide-logging-system

### DIFF
--- a/kubernetes/listener-deployment.yaml.ctmpl
+++ b/kubernetes/listener-deployment.yaml.ctmpl
@@ -21,6 +21,10 @@ spec:
         env:
         - name: listener_config
           value: /etc/secondary-analysis/config.json
+        readinessProbe:
+          httpGet:
+            path: /health
+            port: 8080
       volumes:
       - name: listener-config
         secret:
--- a/lira/api/health.py
+++ b/lira/api/health.py
@@ -2,6 +2,6 @@ import logging
 
 
 def get():
-    logger = logging.getLogger("Lira | {module_path}".format(module_path=__name__))
-    logger.info("Health check request received")
+    logger = logging.getLogger("{module_path}".format(module_path=__name__))
+    logger.debug("Health check request received")
     return dict(status="healthy")
--- a/lira/api/notifications.py
+++ b/lira/api/notifications.py
@@ -6,7 +6,7 @@ from flask import current_app
 from cromwell_tools import cromwell_tools
 from lira import lira_utils
 
-logger = logging.getLogger("Lira | {module_path}".format(module_path=__name__))
+logger = logging.getLogger("{module_path}".format(module_path=__name__))
 
 
 def post(body):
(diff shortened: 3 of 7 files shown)


kubernetes/listener-deployment.yaml.ctmpl
```diff
@@ -21,6 +21,11 @@ spec:
         env:
         - name: listener_config
           value: /etc/secondary-analysis/config.json
+        readinessProbe:
+          httpGet:
+            path: /health
+            port: 8080
+          initialDelaySeconds: 5
```
rexwangcc: What is the difference between `periodSeconds` and `initialDelaySeconds`? According to https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#ConfigureProbes, it seems `periodSeconds` is the parameter that controls the frequency? But since the dev lira's log works as expected now (checks health every 10 seconds), I'm a little bit confused here.

rexwangcc: Besides, it seems the health check path change broke the dev lira's ingress. Based on the solution [here](https://stackoverflow.com/questions/42967763/kubernetes-ingress-gce-keeps-returning-502-error) I deleted the listener ingress and recreated it. It works now.

dshiga:
Hi @rexwangcc, good questions. And it turns out that Google is hitting /health at a high rate like before anyway (seems to be a burst of hits every minute or so). So I reverted the health check change, and instead have revamped logging so that Lira will log at info level by default rather than debug (though we can change that to any other level via configuration) and health checks are only logged at debug level. Let me know what you think!

rexwangcc:
@dshiga Personally I still think `/health` is a better place than `/`. It seems `healthz` and `health` are widely used for health check paths. Although for Lira it is ok to use `/`, we might want to follow the conventions unless it is hard to add a `/health` endpoint. (For Cromwell, @hjfbynara is using `/engine/v1/stats`)

Besides, revamping logging levels sounds good to me.

dshiga:
jenkins retest

dshiga:
Okay, I have added back the readiness probe at /health.

rexwangcc (approved):
LGTM! Thanks for the changes here!

rexwangcc:
jenkins retest

rexwangcc:
Good to merge now

dshiga:
jenkins retest

dshiga:
jenkins retest

PR #59, opened 2018-02-23
Allow lira to use cromwell-as-a-service

samanehsan:
Use cromwell-as-a-service to run workflows if "use_caas" and a "caas_key" are specified in the config file.

### DIFF
--- a/kubernetes/listener-deployment-setup.sh
+++ b/kubernetes/listener-deployment-setup.sh
@@ -5,7 +5,6 @@
 #   - Update the dss storage service white list (https://github.com/HumanCellAtlas/data-store/blob/master/environment#L63)
 #     to include bluebox-subscription-manager@<GCLOUD_PROJECT>.iam.gserviceaccount.com
 
-
 ENV=$1  #name of the deployment environment (e.g. staging)
 GCLOUD_PROJECT=$2  #ID of the gcloud project (e.g. broad-dsde-mint-staging)
 BLUEBOX_SUBSCRIPTION_KEY_DIR=$3  #absolute path of the subscription key
@@ -15,7 +14,13 @@ SS2_PREFIX=$6
 TENX_PREFIX=$7
 LOG_SINK_NAME=$8
 LOG_DESTINATION=$9
+USE_CAAS=${10}
+COLLECTION_NAME=${11:-""}
 VAULT_TOKEN_FILE=${VAULT_TOKEN_FILE:-"$HOME/.vault-token"}
+CROMWELL_URL=https://cromwell.mint-$ENV.broadinstitute.org/api/workflows/v1
+if ${USE_CAAS}; then
+    CROMWELL_URL=https://cromwell.caas-dev.broadinstitute.org/api/workflows/v1
+fi
 
 error=0
 if [ -z $ENV ]; then
@@ -54,9 +59,16 @@ if [ -z $LOG_DESTINATION ]; then
     echo -e "\nYou must specify a log destination of form pubsub.googleapis.com/projects/<logs_project_id>/topics/<topic_name>"
     error=1
 fi
-
+if [ -z $USE_CAAS ]; then
+    echo -e "\nYou must specify whether to use Cromwell-as-a-Service"
+    error=1
+fi
+if [ $USE_CAAS ] && [ -z $COLLECTION_NAME ]; then
+    echo -e "\nYou must specify a collection to use for creating workflows in Cromwell-as-a-Service"
+    error=1
+fi
 if [ $error -eq 1 ]; then
-    echo -e "\nUsage: bash listener-deployment.sh ENV GCLOUD_PROJECT BLUBOX_SUBSCRIPTION_KEY_DIR DSS_URL PIPELINE_TOOLS_PREFIX SS2_PREFIX TENX_PREFIX LOG_SINK_NAME LOG_DESTINATION\n"
+    echo -e "\nUsage: bash listener-deployment.sh ENV GCLOUD_PROJECT BLUBOX_SUBSCRIPTION_KEY_DIR DSS_URL PIPELINE_TOOLS_PREFIX SS2_PREFIX TENX_PREFIX LOG_SINK_NAME LOG_DESTINATION USE_CAAS COLLECTION_NAME\n"
     exit 1
 fi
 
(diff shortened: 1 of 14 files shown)


requirements.txt
```diff
@@ -5,5 +5,6 @@ google-cloud-storage
 flask
 gunicorn
 requests
-git+git://github.com/broadinstitute/cromwell-tools.git
+oauth2client
+git+git://github.com/broadinstitute/cromwell-tools.git@a939745db83fd4b5c72430104592ec39c11cb600
```
samanehsan: I temporarily changed this to the commit from the cromwell-tools PR that includes using caas for testing. I'll change it back before merging!

samanehsan:
jenkins retest

kubernetes/listener-deployment-setup.sh
```diff
@@ -15,7 +14,13 @@ SS2_PREFIX=$6
 TENX_PREFIX=$7
 LOG_SINK_NAME=$8
 LOG_DESTINATION=$9
+USE_CAAS=${10}
```
rexwangcc, line 17: Why is `${10}` here rather `$10`?

kubernetes/listener-deployment-setup.sh
```diff
@@ -15,7 +14,13 @@ SS2_PREFIX=$6
...
+COLLECTION_NAME=${11:-""}
 VAULT_TOKEN_FILE=${VAULT_TOKEN_FILE:-"$HOME/.vault-token"}
+CROMWELL_URL=https://cromwell.mint-$ENV.broadinstitute.org/api/workflows/v1
+if ${USE_CAAS}; then
+    CROMWELL_URL=https://cromwell.caas-dev.broadinstitute.org/api/workflows/v1
```
rexwangcc, line 22: So if we use CAAS here, currently the `caas-dev` is the only option for us?

kubernetes/listener-deployment.sh
```diff
@@ -10,7 +10,12 @@ CERT_DIR=$3
 FULLCHAIN_FILENAME=$4
 PRIVKEY_FILENAME=$5
 DNS_ZONE_NAME=$6
-DOCKER_TAG=$7 #version of the quay.io/humancellatlas/secondary-analysis-lira image to deploy
+DOCKER_TAG=$7 # version of the quay.io/humancellatlas/secondary-analysis-lira image to deploy
+GCLOUD_PROJECT=$8
+USE_CAAS=$9  # whether or not to use cromwell-as-a-service
+WORKFLOW_COLLECTION_ID=${10}  # The name of the workflow-collection to create in SAM (e.g. "dev-workflows")
```
rexwangcc, line 16: What is the difference between `$10` and `${10}`?

kubernetes/listener-deployment.sh
```diff
@@ -110,13 +124,21 @@ aws route53 change-resource-record-sets \
...
 
+
 # Set listener config secret
 printf "\nset listener config secret"
-LIRA_CONFIG_PATH=$CERT_DIR/config.json
-CONFIG_SECRET_NAME=$(bash populate-listener-config-secret.sh $LIRA_CONFIG_PATH)
+CONFIG_FILE=$CERT_DIR/config.json
+
+if $USE_CAAS; then
+    bash caas_login.py ${GCLOUD_PROJECT} ${ENV} ${CERT_DIR} ${WORKFLOW_COLLECTION_ID} ${FIRECLOUD_CONTACT_EMAIL} ${FIRECLOUD_GROUP_NAME}
```
rexwangcc: I guess `caas_login.py` here was a bash script?

kubernetes/populate-listener-config-secret.sh
```diff
@@ -11,6 +12,8 @@ fi
...
 echo $config_secret_name
 
-kubectl create secret generic \
-    $config_secret_name \
-    --from-file=config=$config_file \
+if [ -z $caas_key_file ]; then
+    kubectl create secret generic $config_secret_name --from-file=config=$config_file
+else
+    kubectl create secret generic $config_secret_name --from-file=config=$config_file --from-file=caas_key=$caas_key_file
+fi
```
rexwangcc: nit: new line at the end

lira/api/notifications.py
```diff
@@ -51,15 +52,26 @@ def post(body):
         }
         status_code = 201
     else:
+        # Get Cromwell auth
+        if lira_config.use_caas:
+            auth = {
+                'caas_key': os.environ.get('caas_key'),
```
rexwangcc: I'm wondering if we could move this to `lira.py`, to let Lira choose between Cromwell instances at the startup. As long as lira started, I don't think there is a chance we will change the Cromwell instance from a non-caas to a caas one? I feel like using`os.environ.get('')` in endpoint controllers could cause some potential problems. What do you think?

rexwangcc: For example, maybe we could `os.environ.get('caas_key')` in the lira.py and store it to the lira_config, and then `lira_config.caas_key` here?

run_docker.sh
```diff
@@ -21,15 +23,31 @@ fi
...
     $port
+else
+    docker run \
+        --name lira \
+        --publish $port:$port \
+        -v $config:$mounted_config:ro \
+        -e listener_config=$mounted_config \
+        lira:$tag \
+        $port
+fi
```
rexwangcc: nit: new line at the end

rexwangcc:
Dropped a few comments. Generally this looks great to me!

kubernetes/listener-deployment-setup.sh
samanehsan: Yes, caas is only available in dev at the moment

kubernetes/listener-deployment-setup.sh
samanehsan: This is necessary for positional parameters with more than one digit

kubernetes/listener-deployment.sh
samanehsan: Good catch! This should be `caas_login.sh`: https://github.com/HumanCellAtlas/lira/blob/master/kubernetes/caas_login.sh

samanehsan:
@rexwangcc I updated this PR to pass in CAAS-specific options to the workflows. Could you please take another look?

rexwangcc (approved):
LGTM!

rexwangcc (approved):
👍

PR #60, opened 2018-03-07
Add smartseq2 query for v5 metadata

samanehsan:
Add query for smartseq2 bundles that follows the v5 metadata schema: https://github.com/HumanCellAtlas/metadata-schema/tree/develop

### DIFF
--- a/scripts/v5_queries/smartseq2-query.json
+++ b/scripts/v5_queries/smartseq2-query.json
@@ -1,0 +1,57 @@
+{
+  "query": {
+    "bool": {
+      "must": [
+        {
+          "match": {
+            "files.process_json.processes.content.library_construction_approach": "Smart-seq2"
+
+          }
+        },
+        {
+          "match": {
+            "files.process_json.processes.content.paired_ends": true
+          }
+        },
+        {
+          "match": {
+            "files.biomaterial_json.biomaterials.content.biomaterial_core.ncbi_taxon_id": 9606
+          }
+        }
+      ],
+      "should": [
+        {
+          "match": {
+            "files.process_json.processes.content.biomaterial_collection.dissociation_process.dissociation_method": "FACS"
+          }
+        },
+        {
+          "match": {
+            "files.process_json.processes.content.biomaterial_collection.dissociation_process.dissociation_method": "mouth pipette"
+          }
+        }
+      ],
+      "must_not": [
+        {
+          "match": {
+            "files.process_json.processes.content.process_type.text": "analysis"
+          }
+        }, 
+        {
+          "range": {
+            "files.biomaterial_json.biomaterials.content.biomaterial_core.ncbi_taxon_id": {
+              "lt": 9606
+            }
+          }
+        },
+        {
+          "range": {
+            "files.biomaterial_json.biomaterials.content.biomaterial_core.ncbi_taxon_id": {
+              "gt": 9606
+            }
+          }
+        }
+      ]
+    }
+  }
+}


hannes-ucsc:
This looks like it's not the final version we came up with yesterday?

scripts/v5_queries/smartseq2-query.json
```diff
@@ -0,0 +1,43 @@
...
+            "files.process_json.biomaterial_collection.dissociation_process.dissociation_method": "FACS"
+          }
+        },
+        {
+          "match": {
+            "files.process_json.biomaterial_collection.dissociation_process.dissociation_method": "mouth-pipetting"
+          }
+        }
+      ],
+      "must_not": [
```
malloryfreeberg: I don't think this `must_not` will work. But, I need to add an analysis block to the current v5 process bundle so we can test this.

hannes-ucsc (approved):

hannes-ucsc:
Thanks to @mikebaumann for his clever (and impressively clairvoyant) suggestion of using `must-not` and `range`.

scripts/v5_queries/smartseq2-query.json
```diff
@@ -0,0 +1,57 @@
...
+          }
+        },
+        {
+          "match": {
+            "files.process_json.processes.content.paired_ends": true
+          }
+        },
+        {
+          "match": {
+            "files.biomaterial_json.biomaterials.content.biomaterial_core.ncbi_taxon_id": 9606
```
rexwangcc, line 18: I think it is good for us to keep both the `must` and the `must not` for readability.

rexwangcc: Even they are functionally equal to each other.

scripts/v5_queries/smartseq2-query.json
```diff
@@ -0,0 +1,57 @@
...
+        {
+          "match": {
+            "files.process_json.biomaterial_collection.dissociation_process.dissociation_method": "mouth pipette"
+          }
+        }
+      ],
+      "must_not": [
+        {
+          "exists": {
+            "field": "files.analysis_process_json"
```
rexwangcc: Are there any further changes required for this field?

scripts/v5_queries/smartseq2-query.json
```diff
@@ -0,0 +1,57 @@
...
+            "files.process_json.biomaterial_collection.dissociation_process.dissociation_method": "FACS"
+          }
+        },
+        {
+          "match": {
+            "files.process_json.biomaterial_collection.dissociation_process.dissociation_method": "mouth pipette"
+          }
+        }
+      ],
+      "must_not": [
```
malloryfreeberg: I think this must_not exists field needs to be 
`"field": "files.process_json.analysis.analysis_process"`

rexwangcc (approved):
LGTM. Special thanks to @mikebaumann who already provided the perfect solution to filter out human id 4 months ago!

malloryfreeberg:
I think the analysis must_not exists field needs to be:
"field": "files.process_json.analysis.analysis_process"

samanehsan:
Just fixed that field @malloryfreeberg! Thanks!

malloryfreeberg:
Great! There is now an example v5 bundle with an analysis process section here: https://github.com/HumanCellAtlas/metadata-schema/tree/malloryfreeberg-post_analysis_v5_bundle_example/examples/bundles/v5/Q4DemoSS2PostAnalysis

Once it's put in the data store, you can test the query out!

malloryfreeberg (approved):
Looks great!

scripts/v5_queries/smartseq2-query.json
hannes-ucsc: This does work:

```
          {
            "match": {
              "files.process_json.processes.content.process_type.text": "analysis"
            }
          },
```

scripts/v5_queries/smartseq2-query.json
```diff
@@ -0,0 +1,57 @@
...
+        {
+          "match": {
+            "files.biomaterial_json.biomaterials.content.biomaterial_core.ncbi_taxon_id": 9606
+          }
+        }
+      ],
+      "should": [
+        {
+          "match": {
+            "files.process_json.biomaterial_collection.dissociation_process.dissociation_method": "FACS"
```
hannes-ucsc: This doesn't look right. It's missing `.processes.content` after `process_json`

scripts/v5_queries/smartseq2-query.json
```diff
@@ -0,0 +1,57 @@
...
+          }
+        },
+        {
+          "match": {
+            "files.process_json.biomaterial_collection.dissociation_process.dissociation_method": "mouth pipette"
```
hannes-ucsc: ditto

hannes-ucsc (changes requested):

scripts/v5_queries/smartseq2-query.json
malloryfreeberg: Yes, I think this needs to be:
`files.process_json.processes.content.enrichment_method`

samanehsan: Can you explain why the field should be `enrichment_method` @malloryfreeberg? This was originally `files.assay_json.content.single_cell.cell_handling` in the v4 query and that maps to `files.process_json.processes.content.biomaterial_collection.dissociation_process.dissociation_method` according to the field mapping spreadsheet

samanehsan:
@hannes-ucsc I made the changes you requested! This query now matches the example v5 bundle in the data storage service (and not the one with the analysis process section)

hannes-ucsc (approved):

PR #61, opened 2018-03-12
Add 10x query for v5 metadata bundles

samanehsan:
Add query for 10x bundles that follows the v5 metadata schema: https://github.com/HumanCellAtlas/metadata-schema/tree/develop

Here is the query that matches v4.6.1 bundles for reference: https://github.com/HumanCellAtlas/lira/blob/master/scripts/v4_queries/10x-query.json

### DIFF
--- a/scripts/v5_queries/10x-query.json
+++ b/scripts/v5_queries/10x-query.json
@@ -1,0 +1,44 @@
+{
+  "query": {
+    "bool": {
+      "must": [
+        {
+          "match": {
+            "files.process_json.processes.content.biomaterial_collection.dissociation_process.dissociation_method": "10x_v2"
+          }
+        },
+        {
+          "match": {
+            "files.process_json.processes.content.library_construction_approach": "10x_v2"
+          }
+        },
+        {
+          "match": {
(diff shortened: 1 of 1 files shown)


scripts/v5_queries/10x-query.json
```diff
@@ -0,0 +1,44 @@
+{
+  "query": {
+    "bool": {
+      "must": [
+        {
+          "match": {
+            "files.process_json.processes.content.biomaterial_collection.dissociation_process.dissociation_method": "10x_v2"
```
rexwangcc, line 7: Is this the v5 alternative of `files.assay_json.content.single_cell.cell_handling`?

rexwangcc (approved):
A tiny comment, otherwise LGTM.

scripts/v5_queries/10x-query.json
malloryfreeberg: You are correct, @rexwangcc. This field is meant to describe how single cells are separated from each other. In v4.6.1 this was called "cell_handling" but we have standardized and renamed this to "dissociation_method".

malloryfreeberg (approved):
Looks good!

PR #62, opened 2018-03-15
Cw fix cache issue 594

rexwangcc:
Recently we have experienced an issue that only first submission from Lira to Cromwell will succeed. Cromwell will complain that the dependency zip file is empty for all subsequent submissions. Unfortunately, the root cause of the issue is still vague, one possible reason is that the `io.BytesIO` buffer in memory cannot work well with the `lru_cache` and the POST requests (to Cromwell), which means it somehow get "closed" after the first access and cannot be accessible by `requests.post` further. 

This PR fixes the cache issue by forcing the Lira to create the dependencies zip file on the fly for each submission, this works for now and seems to be stable in the future. 

Besides, this PR also lints and moves the caching logic from the controller file to lira_utils, to keep a set of clean and clear controller files.


start_lira.sh
```diff
@@ -10,4 +10,4 @@ if [ -z $port ]; then
     port=8080
 fi
 
-gunicorn lira.lira -b 0.0.0.0:$port
+gunicorn lira.lira:app -b 0.0.0.0:$port
```
dshiga, line 13: Why this change? I think gunicorn looks for "app" by default.

lira/lira.py
```diff
@@ -11,15 +11,9 @@
 from connexion.resolver import RestyResolver
 import argparse
 
+from . import lira_utils
 from . import lira_config
-from .api import notifications
 
-class MaxLevelFilter(object):
```
dshiga, line 17: good catch

dshiga (approved):

start_lira.sh
rexwangcc: Thanks for catching this, I changed here but forgot to remove `application = app.app` in lira.py. Since I would remove the `application = app.app`, I wanted to add `:app` to provide an explicit entry point for gunicorn (and for readers)

PR #63, opened 2018-03-20
Make it possible to use stub submit wdl

dshiga:
See https://elastc.com/c/4btDIPv3/615-stub-submission-wdl

It needs the same name as the regular submit wdl so that it will work with existing adapter wdls without changes. (Existing adapter wdls import a file named submit.wdl.)


rexwangcc (approved):
LGTM!

PR #64, opened 2018-03-20
update smart seq 2 analysis wdls in lira config

jsotobroad:
With the style guide changes in skylab, we need to change the names of the wdls used by the smart seq 2 pipeline


jsotobroad:
changes needed because of - https://github.com/HumanCellAtlas/skylab/pull/92

PR #65, opened 2018-03-22
Enable Lira to add labels to workflows at start

rexwangcc:
This PR take advantages of another PR https://github.com/broadinstitute/cromwell-tools/pull/9 to enable Lira to add labels to workflows at the start. All of the labels should follow the Cromwell's rules, or the workflow will fail to start.

~~**Note:** to add what specific labels to workflow is not determined yet, this PR only focuses on adding the feature, determining sets of labels is out of scope.~~

### DIFF
--- a/lira/api/notifications.py
+++ b/lira/api/notifications.py
@@ -39,9 +39,12 @@ def post(body):
     cromwell_inputs_file = json.dumps(inputs)
 
     # Prepare labels
-    cromwell_labels = json.dumps(body.get('labels'))
-    if cromwell_labels:
-        logger.info("Will add labels: {labels} to the workflow after syntax validation".format(labels=cromwell_labels))
+    labels_from_notification = body.get('labels')
+    cromwell_labels = lira_utils.compose_labels(wdl_config.workflow_name, wdl_config.wdl_version, uuid, version,
+                                                labels_from_notification)
+    cromwell_labels_file = json.dumps(cromwell_labels)
+
+    logger.debug("Added labels {labels} to workflow".format(labels=cromwell_labels_file))
 
     cromwell_submission = current_app.prepare_submission(wdl_config, lira_config.submit_wdl)
     logger.info(current_app.prepare_submission.cache_info())
@@ -64,7 +67,8 @@ def post(body):
             url=lira_config.cromwell_url,
             user=lira_config.cromwell_user,
             password=lira_config.cromwell_password,
-            label=cromwell_labels
+            label=cromwell_labels_file,
+            validate_labels=False  # switch off the validators provided by cromwell_tools
         )
         if cromwell_response.status_code > 201:
             logger.error("HTTP error content: {content}".format(content=cromwell_response.text))
--- a/lira/api/notifications.py
+++ b/lira/api/notifications.py
@@ -38,6 +38,11 @@ def post(body):
     inputs = lira_utils.compose_inputs(wdl_config.workflow_name, uuid, version, lira_config.env)
     cromwell_inputs_file = json.dumps(inputs)
 
+    # Prepare labels
+    cromwell_labels = json.dumps(body.get('labels'))
+    if cromwell_labels:
+        logger.info("Will add labels: {labels} to the workflow after syntax validation".format(labels=cromwell_labels))
+
     cromwell_submission = current_app.prepare_submission(wdl_config, lira_config.submit_wdl)
     logger.info(current_app.prepare_submission.cache_info())
 
@@ -58,7 +63,8 @@ def post(body):
             options_file=cromwell_submission.options_file,
             url=lira_config.cromwell_url,
             user=lira_config.cromwell_user,
-            password=lira_config.cromwell_password
+            password=lira_config.cromwell_password,
+            label=cromwell_labels
         )
         if cromwell_response.status_code > 201:
             logger.error("HTTP error content: {content}".format(content=cromwell_response.text))
--- a/lira/lira_config.py
+++ b/lira/lira_config.py
@@ -2,6 +2,7 @@
 """
 import logging
 import sys
+from . import lira_utils
 
 
 class Config(object):
@@ -15,7 +16,7 @@ class Config(object):
 
         :param dict config_dictionary: configuration values to be verified
         :param flask_config_values: flask.config.Config configuration values, for example
-          those generated from a connexxion App
+          those generated from a connexion App
         """
         self.config_dictionary = config_dictionary
 
@@ -85,7 +86,8 @@ class WdlConfig(Config):
             'analysis_wdls',
             'workflow_name',
             'wdl_static_inputs_link',
-            'options_link'
+            'options_link',
+            'wdl_version'
         }
 
     def _verify_fields(self):
@@ -94,9 +96,14 @@ class WdlConfig(Config):
             raise TypeError('analysis_wdls must be a list')
 
     def __str__(self):
-        s = 'WdlConfig({0}, {1}, {2}, {3}, {4}, {5})'
+        s = 'WdlConfig({0}, {1}, {2}, {3}, {4}, {5}, {6})'
         return s.format(self.subscription_id, self.wdl_link, self.analysis_wdls,
-            self.workflow_name, self.wdl_static_inputs_link, self.options_link)
+            self.workflow_name, self.wdl_static_inputs_link, self.options_link, self.wdl_version)
+
+    def __repr__(self):
+        s = 'WdlConfig({0}, {1}, {2}, {3}, {4}, {5}, {6})'
+        return s.format(self.subscription_id, self.wdl_link, self.analysis_wdls,
+            self.workflow_name, self.wdl_static_inputs_link, self.options_link, self.wdl_version)
 
 
 class LiraConfig(Config):
@@ -143,7 +150,9 @@ class LiraConfig(Config):
         wdl_configs = []
         try:
             for wdl in config_object['wdls']:
-                wdl_configs.append(WdlConfig(wdl))
+                wdl_configs.append(WdlConfig(lira_utils.merge_two_dicts(wdl,
+                    {'wdl_version': lira_utils.parse_github_resource_url(wdl['analysis_wdls'][0]).version})
+                ))
         except KeyError:
             raise ValueError('supplied config file must contain a "wdls" section.')
         self._verify_wdl_configs(wdl_configs)
@@ -182,6 +191,12 @@ class LiraConfig(Config):
                 'contents.')
 
     def __str__(self):
+        s = 'LiraConfig({0}, {1}, {2}, {3}, {4}, {5}, {6}, {7})'
+        return s.format(self.env, self.submit_wdl, self.cromwell_url,
+            '(cromwell_user)', '(cromwell_password)', '(notification_token)',
+            self.MAX_CONTENT_LENGTH, self.wdls)
+
+    def __repr__(self):
         s = 'LiraConfig({0}, {1}, {2}, {3}, {4}, {5}, {6}, {7})'
         return s.format(self.env, self.submit_wdl, self.cromwell_url,
             '(cromwell_user)', '(cromwell_password)', '(notification_token)',
--- a/lira/lira_config.py
+++ b/lira/lira_config.py
@@ -96,14 +96,32 @@ class WdlConfig(Config):
             raise TypeError('analysis_wdls must be a list')
 
     def __str__(self):
-        s = 'WdlConfig({0}, {1}, {2}, {3}, {4}, {5}, {6})'
-        return s.format(self.subscription_id, self.wdl_link, self.analysis_wdls,
-            self.workflow_name, self.wdl_static_inputs_link, self.options_link, self.wdl_version)
+        s = 'WdlConfig(subscription_id: {0},' \
+            ' wdl_link: {1},' \
+            ' analysis_wdls: {2},' \
+            ' workflow_name: {3},' \
+            ' wdl_static_inputs_link: {4},' \
+            ' options_link: {5},' \
+            ' wdl_version: {6})'
+        return s.format(
+            self.subscription_id,
+            self.wdl_link,
+            self.analysis_wdls,
+            self.workflow_name,
+            self.wdl_static_inputs_link,
+            self.options_link,
+            self.wdl_version)
 
     def __repr__(self):
         s = 'WdlConfig({0}, {1}, {2}, {3}, {4}, {5}, {6})'
-        return s.format(self.subscription_id, self.wdl_link, self.analysis_wdls,
-            self.workflow_name, self.wdl_static_inputs_link, self.options_link, self.wdl_version)
+        return s.format(
+            self.subscription_id,
+            self.wdl_link,
+            self.analysis_wdls,
+            self.workflow_name,
+            self.wdl_static_inputs_link,
+            self.options_link,
+            self.wdl_version)
 
 
 class LiraConfig(Config):
@@ -149,14 +167,14 @@ class LiraConfig(Config):
         # parse the wdls section
         wdl_configs = []
         try:
-            for wdl in config_object['wdls']:
+            for wdl in config_object['wdls']:  # Parse wdls and instantiate WdlConfig objects
                 wdl_configs.append(WdlConfig(lira_utils.merge_two_dicts(wdl,
                     {'wdl_version': lira_utils.parse_github_resource_url(wdl['analysis_wdls'][0]).version})
-                ))
+                ))  # Add wdl_version to wdlConfig objects by parsing each of the first URL of analysis_wdls
         except KeyError:
             raise ValueError('supplied config file must contain a "wdls" section.')
         self._verify_wdl_configs(wdl_configs)
-        config_object['wdls'] = wdl_configs
+        config_object['wdls'] = wdl_configs  # Store the WdlConfig objects back to wdls section
 
         if config_object.get('dry_run'):
             logger = logging.getLogger('{module_path}'.format(module_path=__name__))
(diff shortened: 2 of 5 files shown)


lira/api/notifications.py
```diff
@@ -38,6 +38,11 @@ def post(body):
     inputs = lira_utils.compose_inputs(wdl_config.workflow_name, uuid, version, lira_config.env)
     cromwell_inputs_file = json.dumps(inputs)
 
+    # Prepare labels
+    cromwell_labels = json.dumps(body.get('labels'))
```
dshiga: This is fine for testing purposes but we should change it because blue box won't send labels in the notification. I think we should default to adding bundle id, bundle version, workflow name, workflow version, and pipeline tools version. These can all be obtained from either the notification or the wdl config (some parsing of wdl paths is required). For efficiency, we might want to precompute those that require parsing and add them to the wdl config.

dshiga: I forgot this is a WIP, I'm sure you're thinking about these things already! Anyway, those are my two cents.

lira/lira_config.py
```diff
@@ -94,9 +96,14 @@ def _verify_fields(self):
             raise TypeError('analysis_wdls must be a list')
 
     def __str__(self):
-        s = 'WdlConfig({0}, {1}, {2}, {3}, {4}, {5})'
+        s = 'WdlConfig({0}, {1}, {2}, {3}, {4}, {5}, {6})'
         return s.format(self.subscription_id, self.wdl_link, self.analysis_wdls,
-            self.workflow_name, self.wdl_static_inputs_link, self.options_link)
+            self.workflow_name, self.wdl_static_inputs_link, self.options_link, self.wdl_version)
+
+    def __repr__(self):
```
rexwangcc: implement `__repr__` so that objects are printable in an iterable.

lira/lira_utils.py
```diff
@@ -62,6 +65,66 @@ def compose_inputs(workflow_name, uuid, version, env):
...
+def merge_two_dicts(x, y):
+    """Merge two dictionaries and return the merged result dictionary.
+
+    :param dict x: Input dictionary one.
+    :param dict y: Input dictionary two.
+
+    :return dict: Merged dictionary.
+    """
+    try:
+        return {**x, **y}  # Simple expression in Python3.5+
```
rexwangcc: One of the points makes me like Py3.5+

lira/lira_utils.py
```diff
@@ -62,6 +65,66 @@ def compose_inputs(workflow_name, uuid, version, env):
...
+
+    Note: This is a very temporary solution, once Cromwell is more permissive on labels, this function should be
+     deprecated right away. This function will convert integers to strings, all Upper letters to lower letters,
+     replace all '_' to '-', and replace all '.' to literal 'dot'.
+
+    :param str|int label: Any value of label need to be legalized.
+
+    :return str: A converted, uglified but valid version of label.
+    """
+    return str(label).lower().replace('_', '-').replace('.', 'dot')
```
rexwangcc: This is not elegant at all. However, having this function here enables us to add `bundle versions` and `workflow versions` from now. Once Cromwell is more permissive on labels, we just need to deprecate this function, no need to add any code. Personally, I prefer this way rather than ignoring `bundle versions` and `workflow versions` and add them in the future.

lira/api/notifications.py
```diff
@@ -38,6 +38,19 @@ def post(body):
     inputs = lira_utils.compose_inputs(wdl_config.workflow_name, uuid, version, lira_config.env)
     cromwell_inputs_file = json.dumps(inputs)
 
+    # Prepare labels
+    cromwell_labels = {
+        "bundle-uuid": lira_utils.legalize_cromwell_labels(uuid),
```
rexwangcc: This could possibly slow down the response time of lira, but will be removed once Cromwell is more permissive.

lira/api/notifications.py
```diff
@@ -38,6 +38,19 @@ def post(body):
...
+        "bundle-version": lira_utils.legalize_cromwell_labels(version),
+        "workflow-name": lira_utils.legalize_cromwell_labels(wdl_config.workflow_name),
+        "workflow-version": lira_utils.legalize_cromwell_labels(wdl_config.wdl_version)
+    }
+
+    labels_from_notification = json.dumps(body.get('labels'))
+    if labels_from_notification:
```
rexwangcc: This enables us to add labels in notifications, which is great for testing purposes.

lira/lira_config.py
```diff
@@ -143,7 +150,9 @@ def __init__(self, config_object, *args, **kwargs):
         wdl_configs = []
         try:
             for wdl in config_object['wdls']:
-                wdl_configs.append(WdlConfig(wdl))
+                wdl_configs.append(WdlConfig(lira_utils.merge_two_dicts(wdl,
+                    {'wdl_version': lira_utils.parse_github_resource_url(wdl['analysis_wdls'][0]).version})
+                ))
```
rexwangcc: Parse and add `'wdl_version'` when it is creating the `wdls` properties of lira_config objects.

lira/lira_utils.py
```diff
@@ -62,6 +65,66 @@ def compose_inputs(workflow_name, uuid, version, env):
...
+    if url.startswith('git@') or url.endswith('.git'):
+        raise ValueError('{} is not a valid url to a resource file on Github.'.format(url))
+
+    ParseResult = namedtuple('ParseResult', 'scheme netloc owner repo version path file')
+
+    intermediate_result = urlparse(url)
+    scheme, netloc, path_array = intermediate_result.scheme, intermediate_result.netloc,\
+                                 intermediate_result.path.split('/')
+
+    if netloc == 'github.com':
```
rexwangcc, line 94: This may not be necessary here(lira won't give it this kind of urls), but it is still here, only for "completeness of development" purpose.

lira/api/notifications.py
```diff
@@ -38,6 +38,19 @@ def post(body):
...
+        cromwell_labels.update(labels_from_notification)
+    logger.info("Will add labels: {labels} to the workflow".format(labels=cromwell_labels))
```
dshiga: I think this should be at debug level instead of info.

lira/lira_utils.py
dshiga: Does the Python2 version also work in Python3? If so, I think we should just do the Python2 version to reduce lines of code.

lira/api/notifications.py
```diff
@@ -38,6 +38,19 @@ def post(body):
     inputs = lira_utils.compose_inputs(wdl_config.workflow_name, uuid, version, lira_config.env)
     cromwell_inputs_file = json.dumps(inputs)
 
+    # Prepare labels
+    cromwell_labels = {
```
dshiga: I think this would be a good thing to put into its own function, to reduce the length of the notifications function and also make the prepare labels functionality independently testable.

lira/test/test_lira_utils.py
```diff
@@ -5,6 +5,12 @@
 
 class TestUtils(unittest.TestCase):
 
+    @classmethod
+    def setUpClass(cls):
+        cls.valid_github_url = 'https://github.com/Organization/Repo/blob/master/lib/utils/util.py'
+        cls.valid_github_raw_url = 'https://raw.githubusercontent.com/User1/Repo/v0.3.4/lib/utils/util.py'
```
dshiga: Let's test with a real skylab or pipeline-tools url or both. Might as well make the test data as realistic as possible.

lira/lira_utils.py
```diff
@@ -62,6 +65,66 @@ def compose_inputs(workflow_name, uuid, version, env):
...
+
+    :raises ValueError: Raise a ValueError when the input URL is invalid.
+    """
+    if url.startswith('git@') or url.endswith('.git'):
+        raise ValueError('{} is not a valid url to a resource file on Github.'.format(url))
+
+    ParseResult = namedtuple('ParseResult', 'scheme netloc owner repo version path file')
+
+    intermediate_result = urlparse(url)
+    scheme, netloc, path_array = intermediate_result.scheme, intermediate_result.netloc,\
```
dshiga, line 91: Could we use a library for this instead? https://docs.python.org/2/library/urlparse.html

lira/api/notifications.py
ambrosejcarr: I would write this in past tense instead of future. 

```python
Added labels {labels} to workflow".format(labels=cromwell_labels)
```

lira/lira_utils.py
dshiga: Oh sorry, I see you are using it!

dshiga (commented):
I made a few suggestions.

lira/lira_config.py
```diff
@@ -94,9 +96,14 @@ def _verify_fields(self):
             raise TypeError('analysis_wdls must be a list')
 
     def __str__(self):
-        s = 'WdlConfig({0}, {1}, {2}, {3}, {4}, {5})'
+        s = 'WdlConfig({0}, {1}, {2}, {3}, {4}, {5}, {6})'
```
ambrosejcarr: To make this a bit more readable, could you either: 

1. use keywords instead of numbers, or
2. multi-line the format string so you can more easily read the order of what is going on in the WdlConfig str: 

```python
s = 'WdlConfig({0}, {1}, {2}, {3}, {4}, {5}, {6})'
return s.format(
    self.subscription_id, 
    self.wdl_link, 
    self.analysis_wdls,
    self.workflow_name, 
    self.wdl_static_inputs_link, 
    self.options_link
```

lira/lira_config.py
ambrosejcarr: This should just return `self.__str__()`, since the output is the same. 👍

lira/lira_config.py
```diff
@@ -143,7 +150,9 @@ def __init__(self, config_object, *args, **kwargs):
         wdl_configs = []
         try:
             for wdl in config_object['wdls']:
-                wdl_configs.append(WdlConfig(wdl))
+                wdl_configs.append(WdlConfig(lira_utils.merge_two_dicts(wdl,
```
ambrosejcarr: I think this could use an in-text comment.

lira/lira_config.py
```diff
@@ -187,6 +196,12 @@ def __str__(self):
             '(cromwell_user)', '(cromwell_password)', '(notification_token)',
             self.MAX_CONTENT_LENGTH, self.wdls)
 
+    def __repr__(self):
```
ambrosejcarr: multi-line the `__str__` command as requested for `WdlConfig`, then have `__repr__` return `self.__str__()`.

lira/lira_utils.py
```diff
@@ -62,6 +65,66 @@ def compose_inputs(workflow_name, uuid, version, env):
     }
 
 
+def parse_github_resource_url(url):
+    """Parse a URL which describes a resource file on Github.
+
+    :param str url: A valid URL which describes a resource file on Github.
```
ambrosejcarr: what makes a URL valid? Probably good to tell the user.

lira/lira_utils.py
```diff
@@ -62,6 +65,66 @@ def compose_inputs(workflow_name, uuid, version, env):
...
+
+    intermediate_result = urlparse(url)
+    scheme, netloc, path_array = intermediate_result.scheme, intermediate_result.netloc,\
+                                 intermediate_result.path.split('/')
+
+    if netloc == 'github.com':
+        owner, repo, version, file = path_array[1], path_array[2], path_array[4], path_array[-1]
+        path = '/'.join(path_array[5:])
+    elif netloc == 'raw.githubusercontent.com':
+        owner, repo, version, file = path_array[1], path_array[2], path_array[3], path_array[-1]
```
ambrosejcarr, line 98: If the contents of path_array are standard across the variable, we could use a `namedtuple` here to store them, which would make it much clearer what the contents are.

lira/lira_utils.py
```diff
@@ -62,6 +65,66 @@ def compose_inputs(workflow_name, uuid, version, env):
...
+
+
+def legalize_cromwell_labels(label):
+    """Legalize invalid labels so that they can be accepted by Cromwell.
+
+    Note: This is a very temporary solution, once Cromwell is more permissive on labels, this function should be
+     deprecated right away. This function will convert integers to strings, all Upper letters to lower letters,
+     replace all '_' to '-', and replace all '.' to literal 'dot'.
+
+    :param str|int label: Any value of label need to be legalized.
```
ambrosejcarr: I might define this as requiring a string, and do the int conversion outside so that the reader can better see when the label might be an int.

lira/api/notifications.py
rexwangcc: That is a great idea, will do!

lira/api/notifications.py
rexwangcc: Will do both!

lira/lira_config.py
rexwangcc: Will do. Besides, I also prefer to modify the `__str__` to:
```python
s = 'WdlConfig(subscription_id: {0}, wdl_link: {1}, analysis_wdls: {2}, workflow_name: {3}, wdl_static_inputs_link: {4}, options_link: {5}, wdl_version: {6})'
return s.format(
    self.subscription_id, 
    self.wdl_link, 
    self.analysis_wdls,
    self.workflow_name, 
    self.wdl_static_inputs_link, 
    self.options_link,
    self.wdl_version
```
to provide a better readbility. And the `__repr__` will remain what it is now.

lira/lira_config.py
rexwangcc: Good catch here! Actually I should have removed `__str__` here since `__repr__` will be called as `__str__` automatically. I'm going to change the output of `__str__` now so it won't be a problem anymore.

lira/lira_config.py
rexwangcc: Could you elaborate here?

lira/lira_utils.py
rexwangcc: Will elaborate, thanks!

lira/lira_utils.py
rexwangcc: It is already returning a namedtuple here, I worry about creating another tuple here will make it more complicated than simpler. What do you think?

ambrosejcarr: I don't see a problem with it. Maybe source another opinion. It'd just be nicer to know exactly what you're taking from path_array.

lira/lira_config.py
ambrosejcarr: I think what I'm looking for is a short expansion of `# parse the wdls section` to include `#parse the wdls section by ... <doing things>`

lira/lira_utils.py
rexwangcc: It works for both versions, will do!

lira/lira_utils.py
rexwangcc: That sounds better, will do!

lira/test/test_lira_utils.py
rexwangcc: That makes more sense to me, thanks and will do!

lira/lira_config.py
rexwangcc: Ah, I see what you mean, will do!

lira/lira_config.py
```diff
@@ -167,14 +167,14 @@ def __init__(self, config_object, *args, **kwargs):
         # parse the wdls section
         wdl_configs = []
         try:
-            for wdl in config_object['wdls']:  # Parse wdls section and instantize WdlConfig objects
+            for wdl in config_object['wdls']:  # Parse wdls and instantize WdlConfig objects
```
dshiga: Typo - should say instantiate here

lira/lira_utils.py
```diff
@@ -112,16 +112,16 @@ def merge_two_dicts(x, y):
     return merged  # Backport compatible in earlier Python versions
 
 
-def legalize_cromwell_labels(label):
-    """Legalize invalid labels so that they can be accepted by Cromwell.
+def legitimize_cromwell_labels(label):
```
dshiga: I think legalize was actually more appropriate, and seems clearer to me.

lira/lira_utils.py
```diff
@@ -112,16 +112,16 @@ def merge_two_dicts(x, y):
...
 
     Note: This is a very temporary solution, once Cromwell is more permissive on labels, this function should be
      deprecated right away. This function will convert integers to strings, all Upper letters to lower letters,
-     replace all '_' to '-', replace all '.' to '-', and replace all ':' to '-' as well.
+     replace all '_' to '-', replace all '.' to '-', and replace all ':' to '-'.
 
     :param str|int label: Any value of label need to be legalized.
 
-    :return str: A converted, uglified but valid version of label.
+    :return str: A converted, uglified but legitimate version of label.
```
dshiga: legal is probably better here

dshiga (commented):
Looks great! I made a few more very minor suggestions. Do you need to update cromwell-tools in the Docker image to fix the failing integration test?

rexwangcc:
retest this please

rexwangcc:
@dshiga @ambrosejcarr I made a few changes to pass the integration test, could you have another look at it? Thanks!

dshiga (approved):
Looks great!

ambrosejcarr:
I'm used to `__str__` being the thing you call when you want to consume output, while `__repr__` is for debugging. Do we have that reversed here? 

[reference](https://www.geeksforgeeks.org/str-vs-repr-in-python/)

rexwangcc:
Good point. I feel like we are on the same page tho. I'm referring to this [post](https://stackoverflow.com/questions/1436703/difference-between-str-and-repr-in-python). I used to `__str__` for readability for users(I think this is similar to `toSting()`in Java, but with slight difference), `__repr__` for debugging(I would interpret "debugging" here as "Mapping anything inside a class to strings but keeping their original forms/shapes as much as possible, for debugging purposes") 

In detail, for a class like:
```python
class num(object):
    def __init__(self, x):
        self.x = x
````
and for `print([num(1), num(2)])`, I would expect:
- With only `__str__`, it gives me `[<__main__.num object at 0x000000000>, <__main__.num object at 0x000000001>]`.
- With only `__repr__`, it gives me `[1, 2]`.
- With both `__str__`, `__repr__`, it gives me `[1, 2]`, but `print(num(1))`will give me a more informative string: `The number is 1`

Besides, with an implementation like:
```python
class num(object):
    def __init__(self, x):
        self.x = x
    def __str__(self):
        return "Number is: " + str(self.x)
    def __repr__(self):
        return str(self.x)
```
Type `num(1)` in console/jupyter notebook will still give us :`1`

@ambrosejcarr what do you think?

ambrosejcarr:
I think you're right, but to be more explicit, with both `__repr__` and `__str__`, in your example, you should get: 

`num(1).__str__()` -> `1`
`num(1).__repr__()` -> `The number is 1`

My main concern is that there is never a point where we pass `repr(thing)` to another function. it's my understand that `repr` should be human-facing while `str` and `bytes` should be machine-facing. If you say "this is how the code works", I'll approve it.

rexwangcc:
Ah, I lied, now I think we have opposite understanding here, since I have:
```python
class num(object):
    def __init__(self, x):
        self.x = x
    def __str__(self):
        return "Number is: " + str(self.x)
    def __repr__(self):
        return str(self.x)
```
and you have:
`num(1).__str__() -> 1`
`num(1).__repr__() -> The number is 1`

Based on [info](https://docs.python.org/3/reference/datamodel.html#object.__repr__),  I think I should make `num(num('1').__repr__())` work at all possibilities, but make '`__str__`' provide convenient or concise representation.

Anyway, this debation also occurs in Django forums, let me just stop it here and merge it lol...

ambrosejcarr (approved):
Ok. Let's be consistent across our code bases going forward, whichever we choose.

rexwangcc:
OK, after revisiting it before merging, I change my mind and would like to choose the other way. Practicalness is better than dogmatism:
> there is never a point where we pass repr(thing) to another function

👍  We'd like to use `str()` naturally but won't use `repr()`.

Will make the changes now and merge it once it passes and tests.

rexwangcc:
Thanks for all the comments. Going to merge it now.

PR #71, opened 2018-04-22
Cw add version endpoint 700

rexwangcc:
- [x] Added or updated tests
- [ ] Updated documentation

Add `/version` endpoint to Lira, with unit tests and updated YAML API definition.

### DIFF
--- a/lira/api/version.py
+++ b/lira/api/version.py
@@ -26,7 +26,7 @@ def get_version():
     version_response = {
         "launch_time": current_app.launch_time,
         "config_version": current_app.config_name,
-        "Lira_version": lira_config.get('version'),
+        "Lira_version": lira_config.version,
         "Cromwell_tools_version": cromwell_tools.__version__ or "Unknown",
         "submit_wdl_version": submit_wdl_version,
         "run_mode": "dry_run" if lira_config.get('dry_run') else "live_run",
--- a/lira/api/version.py
+++ b/lira/api/version.py
@@ -1,0 +1,36 @@
+import logging
+from flask import current_app
+import cromwell_tools
+from lira import lira_utils
+
+
+def get_version():
+    """Gather and return Lira's and all its dependencies' versions."""
+    logger = logging.getLogger("{module_path}".format(module_path=__name__))
+    logger.debug("Version request received")
+
+    lira_config = current_app.config
+
+    workflow_info = {
+        wdl.workflow_name: {
+            'version': wdl.workflow_version,
+            'subscription_id': wdl.subscription_id
+        } for wdl in lira_config.wdls
+    }
+
+    try:
+        submit_wdl_version = lira_utils.parse_github_resource_url(lira_config.get('submit_wdl')).version or "Unknown"
+    except ValueError:
+        submit_wdl_version = "Unknown"
+
+    version_response = {
+        "launch_time": current_app.launch_time,
+        "config_version": current_app.config_name,
+        "Lira_version": lira_config.get('version'),
+        "Cromwell_tools_version": cromwell_tools.__version__ or "Unknown",
+        "submit_wdl_version": submit_wdl_version,
+        "run_mode": "dry_run" if lira_config.get('dry_run') else "live_run",
+        "workflow_info": workflow_info
+    }
+
+    return version_response
--- a/lira/api/version.py
+++ b/lira/api/version.py
@@ -25,9 +25,8 @@ def get_version():
 
     version_response = {
         "launch_time": current_app.launch_time,
-        "config_version": current_app.config_name,
-        "Lira_version": lira_config.version,
-        "Cromwell_tools_version": cromwell_tools.__version__ or "Unknown",
+        "lira_version": lira_config.version,
+        "cromwell_tools_version": cromwell_tools.__version__ or "Unknown",
         "submit_wdl_version": submit_wdl_version,
         "run_mode": "dry_run" if lira_config.get('dry_run') else "live_run",
         "workflow_info": workflow_info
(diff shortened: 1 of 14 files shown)


lira/api/version.py
```diff
@@ -0,0 +1,36 @@
...
+
+    try:
+        submit_wdl_version = lira_utils.parse_github_resource_url(lira_config.get('submit_wdl')).version or "Unknown"
+    except ValueError:
+        submit_wdl_version = "Unknown"
+
+    version_response = {
+        "launch_time": current_app.launch_time,
+        "config_version": current_app.config_name,
+        "Lira_version": lira_config.version,
```
samanehsan: Can we make `Lira_version` and `Cromwell_tools_version` all lowercase for consistency?

lira/lira.yml
```diff
@@ -86,3 +120,18 @@ definitions:
...
       status:
         type: string
+  VersionResponse:
+    type: object
+    properties:
+      launch_time:
+        type: string
+      config_version:
+        type: string
+      Lira_version:
```
samanehsan: Let's rename this to `lira_version` for consistency with the other property names

lira/api/version.py
```diff
@@ -0,0 +1,36 @@
...
+    logger = logging.getLogger("{module_path}".format(module_path=__name__))
+    logger.debug("Version request received")
+
+    lira_config = current_app.config
+
+    workflow_info = {
+        wdl.workflow_name: {
+            'version': wdl.workflow_version,
+            'subscription_id': wdl.subscription_id
+        } for wdl in lira_config.wdls
```
samanehsan, line 18: I forgot python had dictionary comprehensions! Very cool!

lira/lira.yml
samanehsan: Should this include `cromwell_tools_version` here as well?

lira/lira_config.py
```diff
@@ -223,10 +229,17 @@ def _verify_wdl_configs(wdl_configs):
...
-        return s.format(self.env, self.submit_wdl, self.cromwell_url,
-            ('use_caas: ' + self.use_caas), '(notification_token)',
-            self.MAX_CONTENT_LENGTH,self.wdls)
+        s = 'LiraConfig({0}, {1}, {2}, {3}, {4}, {5}, {6}, {7}, {8})'
+        return s.format(
+            self.env,
+            self.submit_wdl,
+            self.cromwell_url,
+            ('use_caas: ' + self.use_caas),
+            '(cromwell_password)',
```
samanehsan: I'm not sure if we want to include `cromwell_password` here since it is optional if lira is using cromwell-as-a-service.

lira/test/test_get_version.py
```diff
@@ -0,0 +1,46 @@
...
+        dir = os.path.abspath(os.path.dirname(__file__))
+        os.chdir(dir)
+        env = mock.patch.dict('os.environ', {'listener_config': 'data/config.json'})
+        with env:
+            # Import lira here to get the config from os.environ correctly
+            from lira import lira
+
+            lira.app.testing = True
+            cls.client = lira.app.app.test_client()
+            lira.app.app.config_name = 'lira-config-2018-01-01-00-59'
```
samanehsan: Are there any side-effects to changing these variables on the lira app? In the job manger we have a test connexion app that the tests use. Could we do something like that here? https://github.com/DataBiosphere/job-manager/blob/master/servers/cromwell/jobs/test/__init__.py#L10
https://github.com/DataBiosphere/job-manager/blob/master/servers/cromwell/jobs/test/test_jobs_controller.py#L23

samanehsan:
This looks great @rexwangcc! I left a few comments on the PR :octocat:

lira/api/version.py
rexwangcc: Sure!

lira/lira.yml
rexwangcc: nice catch! Thanks!

lira/api/version.py
rexwangcc: :P

lira/test/test_get_version.py
rexwangcc: I don't think there are any side-effects here. We import lira here and change the imported lira's app object with this test file's namespace for test cases. AFAK, I can't see this would affect Lira application's functionalities. 

To implement an equivalent to the Job Manager's test case, we need to introduce `flask_testing` module and create a separate `test_requirements.txt` apart from the current `requirements.txt`, which seems overkill and complex to me. I'd prefer not do now that until in the future we want to update our test case completely, (e.g. replace unittest with pytest and separate test requirements from prod requirements). What do you think?

samanehsan: Oh, I see! In that case, I think it's okay to leave it as it is now. Thanks!

lira/lira_config.py
```diff
@@ -236,15 +249,18 @@ def __repr__(self):
             ' cromwell_password(Invisible): {4},' \
             ' notification_token(Invisible): {5},' \
             ' MAX_CONTENT_LENGTH: {6},' \
-            ' wdls: {7})'
+            ' wdls: {7},' \
+            ' Lira_version: {8})'
```
samanehsan: Can you change this to `Lira_version` as well so it's the same as the variable name?

samanehsan (approved):
This looks good to merge! Just one more place that needs a naming change :octocat:

lira/lira_config.py
rexwangcc: Sorry, I totally forgot this place since it's an offline discussion, will do. Thanks!

rexwangcc:
Per our offline discussion, I just removed the `config_name` from the endpoint. @samanehsan Could you have a final look at this? I will move forward to merge it. Thanks!!

rexwangcc:
@samanehsan Please wait for my rebasing...

PR #73, opened 2018-04-30
Fix saving caas key in vault

samanehsan:
Save the service account key for Cromwell-as-a-Service under a key called "value" in the vault path so that the JSON key can be retrieved without the additional wrapper fields that vault adds in.

- [ ] Added or updated tests
- [ ] Updated documentation


rexwangcc (approved):
lgtm!

PR #74, opened 2018-05-08
Make use_caas value boolean

samanehsan:
Remove the quotes around the `USE_CAAS` value in the listener-config template, so that it is a boolean value when it is rendered. Also use `parseBool` to convert "False", "false", 0, etc. to false.

- [ ] Added or updated tests
- [ ] Updated documentation


dshiga (approved):
Looks good!

PR #77, opened 2018-05-24
Add option to include additional metadata in notifications

samanehsan:
Add option to include additional metadata fields in the notification when creating a subscription to the data store. The script will optionally take in a JSON file containing information about which fields to include, where "expression" is using [JMESpath](http://jmespath.org/) syntax to specify the value of the property. Example JSON format from the DSS API:  
```
{
    "additionalProp1": {
      "type": "jmespath", 
      "expression": "string"
    },
    "additionalProp2": {
      "type": "jmespath", 
      "expression": "string"
    }
}
```
This is passed in as the value of the "attachments" key in the JSON payload sent to the `/subscriptions` endpoint. Here is the API spec for additional context: https://github.com/HumanCellAtlas/data-store/blob/1a7a28db83c767aee4da7a93daaea5cad484d8c1/dss-api.yml#L486

The notification body will now have an additional "attachments" key with the additional metadata fields, e.g:
```
    'attachments': {
        'sample_id': ['sample_1'],
        'submitter_id': None, 
        'project_shortname': 'Q4_DEMO-project_PRJNA248302'
    }
```

- [ ] Added or updated tests
- [x] Updated documentation

### DIFF
--- a/scripts/subscription-create.py
+++ b/scripts/subscription-create.py
@@ -6,16 +6,24 @@ from httplib2 import Http
 from oauth2client.service_account import ServiceAccountCredentials
 
 
-def prep_json(callback_base_url, listener_secret, query_json):
-    js = {}
-    js["callback_url"] = "{0}?auth={1}".format(callback_base_url, listener_secret)
-
-    query = None
-    with open(query_json) as f:
+def prep_json(callback_base_url, listener_secret, query_json_file, attachments_file=None):
+    query = {}
+    with open(query_json_file) as f:
         query = json.load(f)
-    js["es_query"] = query
 
-    return js
+    payload = {
+        'es_query': query,
+        'callback_url': '{0}?auth={1}'.format(callback_base_url, listener_secret)
+    }
+
+    if attachments_file:
+        attachments = {}
+        with open(attachments_file) as f:
+            attachments = json.load(f)
+        payload['attachments'] = attachments
+
+    return payload
+
 
 def make_request(js, dss_url, key_file):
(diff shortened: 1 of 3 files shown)


scripts/subscription-create.py
```diff
@@ -6,16 +6,24 @@
...
-    js["es_query"] = query
 
-    return js
+    payload = {
+        'es_query': query,
+        'callback_url': '{0}?auth={1}'.format(callback_base_url, listener_secret)
+    }
+
+    if attachments_file:
+        attachments = {}
```
dshiga: I don't think it's [necessary](https://stackoverflow.com/questions/45100271/scope-of-variable-within-with-statement) to define attachments up here.

scripts/subscription-create.sh
```diff
@@ -5,5 +5,10 @@ green_url=$2
 key_file=$3
 listener_secret=$4
 query_json=$5
+additional_metadata=$6
 
-./subscription-create.py "$dss_url?replica=gcp" $green_url $listener_secret $key_file $query_json
+if [ $additional_metadata ]; then
```
dshiga, line 10: You could avoid duplicating the `subscription-create.py` command by doing something like:
```
if [ $additional_metadata ]; then
  additional_metadata_flag="--additional_metadata"
fi
./subscription-create.py "$dss_url?replica=gcp" $green_url $listener_secret $key_file $query_json $additional_metadata_flag $additional_metadata
```
It's a little less clear though, so up to you.

dshiga (approved):
Looks good!

PR #78, opened 2018-05-30
Support hmac auth

dshiga:
- [x] Added or updated tests
- [x] Updated documentation

https://elastc.com/c/VFwYoDJ$/74-add-hmac-auth-on-blue-side

### DIFF
--- a/kubernetes/caas_login.sh
+++ b/kubernetes/caas_login.sh
@@ -9,7 +9,6 @@ WORKFLOW_COLLECTION_ID=$4  # The name of the workflow-collection to create in SA
 FIRECLOUD_CONTACT_EMAIL=$5  # A contact email to use for the FireCloud service account
 FIRECLOUD_GROUP_NAME=$6  # The name of the user group to create in FireCloud to control workflow collection permissions (e.g. "write-access")
 VAULT_TOKEN_FILE=${VAULT_TOKEN_FILE:-"$HOME/.vault-token"}
-FIRECLOUD_URL=${FIRECLOUD_URL:-"https://firecloud.dsde-dev.broadinstitute.org"}
 FIRECLOUD_API_URL=${FIRECLOUD_API_URL:-"https://firecloud-orchestration.dsde-dev.broadinstitute.org"}
 SAM_URL=${SAM_URL:-"https://sam.dsde-dev.broadinstitute.org"}
 
@@ -31,7 +30,7 @@ git clone git@github.com:broadinstitute/firecloud-tools.git
 cd firecloud-tools
 
 echo "Register the service account for use in Firecloud"
-./run.sh scripts/register_service_account/register_service_account.py -j ${KEY_FILE_PATH} -e ${FIRECLOUD_CONTACT_EMAIL} -u ${FIRECLOUD_URL}
+./run.sh scripts/register_service_account/register_service_account.py -j ${KEY_FILE_PATH} -e ${FIRECLOUD_CONTACT_EMAIL} -u ${FIRECLOUD_API_URL}
 
 cd ..
 
--- a/kubernetes/listener-config.json.ctmpl
+++ b/kubernetes/listener-config.json.ctmpl
@@ -1,9 +1,12 @@
 {{with $environment := env "ENV"}}
 {{with $listenerSecret := vault (printf "secret/dsde/mint/%s/listener/listener_secret" $environment)}}
+{{with $hmac_keys := vault (printf "secret/dsde/mint/%s/listener/hmac_keys" $environment)}}
     {
       {{ if env "USE_CAAS" | parseBool }}
         "collection_name": "{{ env "COLLECTION_NAME" }}",
         "cromwell_url": "https://cromwell.caas-dev.broadinstitute.org/api/workflows/v1",
+        "google_project": "broad-dsde-mint-{{$environment}}",
+        "gcs_root": "gs://broad-dsde-mint-{{$environment}}-cromwell-execution/caas-cromwell-executions",
       {{ else }}
         {{ with $cromwellSecrets := vault (printf "secret/dsde/mint/%s/common/htpasswd" $environment) }}
           "cromwell_password": "{{$cromwellSecrets.Data.cromwell_password}}",
@@ -14,8 +17,14 @@
       "version": "{{ env "LIRA_VERSION" }}",
       "MAX_CONTENT_LENGTH": 10000,
       "use_caas": {{ env "USE_CAAS" | parseBool }},
-      "env": "{{ env "ENV" }}",
+      "env": "{{$environment}}",
+      "dss_url": "https://dss.{{ if eq $environment "test" }}integration{{ else }}{{$environment}}{{ end }}.data.humancellatlas.org/v1",
+      "ingest_url": "http://api.ingest.{{ if eq $environment "test" }}integration{{ else }}{{$environment}}{{ end }}.data.humancellatlas.org/",
+    {{ if env "USE_HMAC" | parseBool }}
+      "hmac_key": "{{$hmac_keys.Data.current_key}}",
+    {{ else }}
       "notification_token": "{{$listenerSecret.Data.notification_token}}",
+    {{ end }}
       "submit_wdl": "{{ env "PIPELINE_TOOLS_PREFIX" }}/adapter_pipelines/{{ env "SUBMIT_WDL_DIR" }}submit.wdl",
       "wdls": [
         {
@@ -45,5 +54,6 @@
         }
       ]
     }
+{{end}}
 {{end}}
 {{end}}
--- a/lira/lira_config.py
+++ b/lira/lira_config.py
@@ -151,7 +151,7 @@ class LiraConfig(Config):
         # Send logs between log_level_lira and info (inclusive) to stdout.
         # If log_level_lira > INFO then nothing is sent to stdout.
         stdout_handler = logging.StreamHandler(sys.stdout)
-        log_level_lira = config_object.get('log_level_lira', None)
+        log_level_lira = config_object.get('log_level_lira')
         if log_level_lira is not None:
             self.log_level_lira = getattr(logging, log_level_lira)
         stdout_handler.setLevel(self.log_level_lira)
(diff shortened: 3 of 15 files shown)


dshiga:
retest this please

dshiga:
retest this please

kubernetes/caas_login.sh
```diff
@@ -9,7 +9,6 @@ WORKFLOW_COLLECTION_ID=$4  # The name of the workflow-collection to create in SA
 FIRECLOUD_CONTACT_EMAIL=$5  # A contact email to use for the FireCloud service account
 FIRECLOUD_GROUP_NAME=$6  # The name of the user group to create in FireCloud to control workflow collection permissions (e.g. "write-access")
 VAULT_TOKEN_FILE=${VAULT_TOKEN_FILE:-"$HOME/.vault-token"}
-FIRECLOUD_URL=${FIRECLOUD_URL:-"https://firecloud.dsde-dev.broadinstitute.org"}
```
rexwangcc, line 12: Could you clarify why we need to change here? Is this a change on the firecloud side?

dshiga: This was a bug in the original script. I tried running register_service_account.py with the firecloud url and it didn't work. It did work when I gave it the orchestration url, which makes sense, because orchestration is the API and firecloud is the UI.

kubernetes/listener-config.json.ctmpl
```diff
@@ -14,8 +17,14 @@
       "version": "{{ env "LIRA_VERSION" }}",
       "MAX_CONTENT_LENGTH": 10000,
       "use_caas": {{ env "USE_CAAS" | parseBool }},
-      "env": "{{ env "ENV" }}",
+      "env": "{{$environment}}",
+      "dss_url": "https://dss.{{ if eq $environment "test" }}integration{{ else }}{{$environment}}{{ end }}.data.humancellatlas.org/v1",
+      "ingest_url": "http://api.ingest.{{ if eq $environment "test" }}integration{{ else }}{{$environment}}{{ end }}.data.humancellatlas.org/",
+    {{ if env "USE_HMAC" | parseBool }}
+      "hmac_key": "{{$hmac_keys.Data.current_key}}",
```
rexwangcc, line 24: It seems `vault read -format=json secret/dsde/mint/dev/listener/hmac_keys | jq .data.current_key` works but `vault read -format=json secret/dsde/mint/dev/listener/hmac_keys | jq .Data.current_key` is not working?

kubernetes/listener-config.json.ctmpl
```diff
@@ -14,8 +17,14 @@
...
+    {{ else }}
       "notification_token": "{{$listenerSecret.Data.notification_token}}",
```
rexwangcc, line 26: Weird, `vault read -format=json secret/dsde/mint/dev/listener/listener_secret | jq .Data.notification_token` is not working, but this file has been here for a while, I'm not sure why this works currently, maybe the ctmpl rendering has a different mechanism from `jq`, say case insensitive?

kubernetes/listener-config.json.ctmpl
dshiga: .Data.current_key does work when rendering the ctmpl file though. We have always used uppercase .Data for the other vault records in the ctmpl file (like cromwell_user on line 13), so I'm following the same pattern here.

kubernetes/listener-config.json.ctmpl
rexwangcc: NVM, just saw your comments above, thanks!

kubernetes/caas_login.sh
rexwangcc: I see, thanks for fixing it!

lira/lira_config.py
```diff
@@ -184,7 +184,11 @@ def __init__(self, config_object, *args, **kwargs):
             if not caas_key:
                 raise ValueError('No service account json key provided for cromwell-as-a-service.')
             self.caas_key = caas_key
-        elif not config_object.get('cromwell_user') and not config_object.get('cromwell_password'):
+            if not config_object.get('google_project'):
+                raise ValueError('No google_project specified')
+            if not config_object.get('gcs_root'):
+                raise ValueError('No gcs_root specified')
```
rexwangcc: Based on the changes to `listener-config.json.ctmpl`, I guess the google project and gcs here refer to the google project that Lira is deployed to. It would be great to have comments here distinguishing this google project and the caas google project.

Besides, just want to confirm, does this mean if Lira is using CaaS as the backend, it can only be deployed to GCP? (rather than AWS)

lira/lira_config.py
```diff
@@ -201,6 +205,13 @@ def __init__(self, config_object, *args, **kwargs):
             logger = logging.getLogger('{module_path}'.format(module_path=__name__))
             logger.warning('***Lira is running in dry_run mode and will NOT launch any workflows***')
 
+        hmac_key = config_object.get('hmac_key')
+        if hmac_key:
+            hmac_key = hmac_key.encode('utf-8')
+            config_object['hmac_key'] = hmac_key
```
rexwangcc: nit: `config_object['hmac_key'] = hmac_key.encode('utf-8')` to save lines (since the variable `hmac_key` is not used anymore, I guess we don't need to update it?)

lira/lira_config.py
```diff
@@ -201,6 +205,13 @@ def __init__(self, config_object, *args, **kwargs):
...
+
+        config_object['stale_notification_timeout'] = config_object.get('stale_notification_timeout', 0)
```
rexwangcc: Could you add comments about what's this parameter for, since this is a separate file from lira_utils.py?

lira/lira_config.py
```diff
@@ -239,8 +251,9 @@ def __str__(self):
             '(notification_token)',
```
rexwangcc: Should this get removed as well?

lira/lira_config.py
```diff
@@ -250,7 +263,9 @@ def __repr__(self):
             ' notification_token(Invisible): {4},' \
```
rexwangcc: Should this get removed as well?

lira/lira_utils.py
```diff
@@ -25,14 +30,99 @@ def response_with_server_header(body, status):
...
+        HTTPSignatureAuth.verify(request, key_resolver=key_resolver)
+
+        # Make sure notification isn't too old
+        _check_date(request.headers.get('Date'), stale_notification_timeout)
+
+        # Make sure given body digest and calculated digest match
+        digest_from_header = request.headers.get('Digest')
+        calculated_digest = _calculate_digest(request)
+        if calculated_digest != digest_from_header:
+            logger.info('Auth error: digests do not match')
```
rexwangcc: I'm not sure if this should be INFO or ERROR?

lira/lira_config.py
dshiga: This is the Google project where the workflow will be run, not necessarily the project that Lira is deployed to (although in our case it is the same project). I added some more text to the error message to make this clearer.

lira/lira_config.py
dshiga: Good point, will fix

lira/lira_config.py
dshiga: will do

lira/lira_utils.py
```diff
@@ -49,45 +139,48 @@ def extract_uuid_version_subscription_id(msg):
...
     options_json = json.loads(options_file)
 
-    with open(caas_key_file) as f:
+    with open(lira_config.caas_key) as f:
         caas_key = f.read()
     options_json.update({
-        'jes_gcs_root': 'gs://broad-dsde-mint-{}-cromwell-execution/caas-cromwell-executions'.format(env),
-        'google_project': 'broad-dsde-mint-{}'.format(env),
+        'jes_gcs_root': lira_config.gcs_root,
+        'google_project': lira_config.google_project,
```
rexwangcc, line 180: nvm my previous comments about GCP, this indicates that we are already going this way.

lira/lira_utils.py
```diff
@@ -25,14 +30,99 @@ def response_with_server_header(body, status):
...
+        # Make sure given body digest and calculated digest match
+        digest_from_header = request.headers.get('Digest')
+        calculated_digest = _calculate_digest(request)
+        if calculated_digest != digest_from_header:
+            logger.info('Auth error: digests do not match')
+            return False
+        return True
+    except KeyError as e:
+        logger.info('Auth error: No HMAC key found in Lira config with id: {0}'.format(e))
+        return False
```
rexwangcc: I'm not sure where in this try block we will get a KeyError?

dshiga: Good catch, we don't raise a KeyError. I'm removing this.

requirements.txt
```diff
@@ -3,11 +3,12 @@ wheel==0.31.0
 connexion==1.4
 google-auth==1.4.1
 google-cloud-storage==1.8.0
-flask==0.12.2
+flask>=1.0.0
```
rexwangcc, line 6: 👍 It is news to me that Flask has v1 now!!

rexwangcc (commented):
I left a few comments, generally, this is an amazing update with sufficient test cases, thanks for doing this!

lira/lira_utils.py
dshiga: Yes, good point, I'll change it to ERROR.

scripts/subscription-create.sh
```diff
@@ -5,10 +5,49 @@ green_url=$2
 key_file=$3
 listener_secret=$4
```
rexwangcc, line 6: What do you think about replacing all `listener_secret ` in this file with `query_param_token ` to avoid confusions?

dshiga: The `listener_secret` param is used to hold either the query param token or the hmac secret, depending on what the user is trying to do. I thought about trying to have separate parameters for them, but it's a bit difficult to support optional parameters in shell scripts.

rexwangcc: I see, then it makes sense to stick with `listener_secret `

rexwangcc (approved):
LGTM! I think maybe we need to coordinate with Blue when we want to deploy this new feature?

dshiga:
Thanks @rexwangcc! Blue box already supports hmac and it can be turned on or off on a per subscription basis, so we can deploy this whenever we want.

PR #79, opened 2018-05-30
Add submit_and_hold flag

samanehsan:
Add flag to configure whether the workflows started by Lira should be set to "On Hold" status by default. By default, the `submit_and_hold` flag is set to false. 

- [ ] Added or updated tests
- [ ] Updated documentation

### DIFF
--- a/lira/api/notifications.py
+++ b/lira/api/notifications.py
@@ -82,6 +82,7 @@ def post(body):
             url=lira_config.cromwell_url,
             label=cromwell_labels_file,
             validate_labels=False,  # switch off the validators provided by cromwell_tools
+            on_hold=lira_config.submit_and_hold,
             **auth
         )
         if cromwell_response.status_code > 201:
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,4 +10,4 @@ requests==2.18.4
 requests-mock==1.4.0
 oauth2client==4.1.2
 mock==2.0.0
-git+git://github.com/broadinstitute/cromwell-tools.git@72f2ac955afdb70b25144ff64f4177d0261deb97
+git+git://github.com/broadinstitute/cromwell-tools.git@v0.4.0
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -10,4 +10,4 @@ requests==2.18.4
 requests-mock==1.4.0
 oauth2client==4.1.2
 mock==2.0.0
-git+git://github.com/broadinstitute/cromwell-tools.git@v0.3.1
+git+git://github.com/broadinstitute/cromwell-tools.git@72f2ac955afdb70b25144ff64f4177d0261deb97
```
samanehsan: I will change this to the new version of cromwell-tools once https://github.com/broadinstitute/cromwell-tools/pull/34 is merged

lira/api/notifications.py
```diff
@@ -82,6 +82,7 @@ def post(body):
             url=lira_config.cromwell_url,
             label=cromwell_labels_file,
             validate_labels=False,  # switch off the validators provided by cromwell_tools
+            on_hold=lira_config.submit_and_hold,
```
rexwangcc, line 85: nit: hold -> onhold so it is consistent with Cromwell's option `workflowOnHold`

rexwangcc (approved):
LGTM! reminder: change the version of cromwell-tools

lira/api/notifications.py
samanehsan: I'm not sure I understand your comment, since the parameter is `on_hold` here. Do you mean change `submit_and_hold` --> `submit_and_on_hold`?

rexwangcc: Sorry for the confusion, yeah that is what I meant, but it is an optional nit.

samanehsan: Ok! I think that will make it a bit verbose, so I would prefer to leave it as-is

PR #81, opened 2018-06-05
Add v6 subscription queries

samanehsan:
Opening up this PR for review, but I would still like to test this query to see if it matches a v6 bundle in the HCA data store before merging.

### DIFF
--- a/scripts/v6_queries/10x-query.json
+++ b/scripts/v6_queries/10x-query.json
@@ -1,0 +1,44 @@
+{
+  "query": {
+    "bool": {
+      "must": [
+        {
+          "match": {
+            "files.protocol_json.protocols.content.dissociation_method.text": "10x_v2"
+          }
+        },
+        {
+          "match": {
+            "files.protocol_json.protocols.content.library_construction_approach.text": "10x_v2"
+          }
+        },
+        {
+          "match": {
+            "files.biomaterial_json.biomaterials.content.biomaterial_core.ncbi_taxon_id": 9606
+          }
+        }
+      ],
+      "must_not": [
+        {
+          "match": {
+            "files.protocol_json.protocols.content.protocol_type.text": "analysis"
+          }
+        },
+        {
+          "range": {
+            "files.biomaterial_json.biomaterials.content.biomaterial_core.ncbi_taxon_id": {
+              "lt": 9606
+            }
+          }
+        },
+        {
+          "range": {
+            "files.biomaterial_json.biomaterials.content.biomaterial_core.ncbi_taxon_id": {
+              "gt": 9606
+            }
+          }
+        }
+      ]
+    }
+  }
+}
--- a/scripts/v6_queries/10x-query.json
+++ b/scripts/v6_queries/10x-query.json
@@ -4,12 +4,12 @@
       "must": [
         {
           "match": {
-            "files.protocol_json.protocols.content.dissociation_method.text": "10x_v2"
+            "files.protocol_json.protocols.content.dissociation_method.text": "10x sequencing"
           }
         },
         {
           "match": {
-            "files.protocol_json.protocols.content.library_construction_approach.text": "10x_v2"
+            "files.protocol_json.protocols.content.library_construction_approach.text": "10x sequencing"
           }
         },
         {
--- a/scripts/v6_queries/metadata_attachments.json
+++ b/scripts/v6_queries/metadata_attachments.json
@@ -1,0 +1,14 @@
+{
+  "project_shortname": {
+    "type": "jmespath",
+    "expression": "files.project_json.content.project_core.project_shortname"
+  },
+  "submitter_id": {
+    "type": "jmespath",
+    "expression": "files.project_json.hca_ingest.submitter_id"
+  },
+  "sample_id": {
+    "type": "jmespath",
+    "expression": "files.links_json.links[?source_type=='biomaterial' && destination_type=='sequencing_process'].source_id"
+  }
+}
--- a/scripts/v6_queries/metadata_attachments.json
+++ b/scripts/v6_queries/metadata_attachments.json
@@ -6,9 +6,5 @@
   "submitter_id": {
     "type": "jmespath",
     "expression": "files.project_json.hca_ingest.submitter_id"
-  },
-  "sample_id": {
-    "type": "jmespath",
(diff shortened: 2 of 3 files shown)


scripts/v6_queries/10x-query.json
```diff
@@ -0,0 +1,44 @@
+{
+  "query": {
+    "bool": {
+      "must": [
+        {
+          "match": {
+            "files.protocol_json.protocols.content.dissociation_method.text": "10x_v2"
+          }
```
malloryfreeberg: `dissociation_method.text": "10X sequencing"`

scripts/v6_queries/10x-query.json
```diff
@@ -0,0 +1,44 @@
...
+        },
+        {
+          "match": {
+            "files.protocol_json.protocols.content.library_construction_approach.text": "10x_v2"
+          }
```
malloryfreeberg: `library_construction_approach.text": "10X sequencing"`

scripts/v6_queries/10x-query.json
```diff
@@ -0,0 +1,44 @@
...
+          "match": {
+            "files.biomaterial_json.biomaterials.content.biomaterial_core.ncbi_taxon_id": 9606
+          }
+        }
+      ],
+      "must_not": [
+        {
+          "match": {
+            "files.protocol_json.protocols.content.protocol_type.text": "analysis"
+          }
```
malloryfreeberg, line 25: `protocol_type` for the `analysis_protocol.json` has changed. @daniwelter what would be the ontology text for Saman to check this?

scripts/v6_queries/10x-query.json
malloryfreeberg: I think we need to extend this ontology to separately capture 10_v2, v1, etc. @daniwelter thoughts?

scripts/v6_queries/10x-query.json
dshiga: Thanks for your help @malloryfreeberg and @daniwelter! For future changes like this, is there a service or repo somewhere that we can use to find the proper terms to use?

scripts/v6_queries/metadata_attachments.json
```diff
@@ -0,0 +1,14 @@
...
+    "type": "jmespath",
+    "expression": "files.project_json.content.project_core.project_shortname"
+  },
+  "submitter_id": {
+    "type": "jmespath",
+    "expression": "files.project_json.hca_ingest.submitter_id"
+  },
+  "sample_id": {
+    "type": "jmespath",
+    "expression": "files.links_json.links[?source_type=='biomaterial' && destination_type=='sequencing_process'].source_id"
```
malloryfreeberg: I think `'sequencing_process'` might need to be changed here, but I'm not sure what the linking will look like with the new bundling proposal. Asking @aaclan-ebi to look at this. 

What is the use of the file? Is it new?

scripts/v6_queries/smartseq2-query.json
```diff
@@ -0,0 +1,57 @@
...
+      ],
+      "should": [
+        {
+          "match": {
+            "files.protocol_json.protocols.content.dissociation_method.text": "fluorescence-activated cell sorting"
+          }
+        },
+        {
+          "match": {
+            "files.protocol_json.protocols.content.dissociation_method.text": "mouth pipette"
```
malloryfreeberg, line 30: I don't think "mouth pipette" is a value term under the ontology governing this field, but also I think "mouth pipette" was chosen as a field to query for for the integration test spreadsheet. We might want to consider deprecating this match in favor of matches to either the project_shortname or some sort of flag field like some people had been discussing a while ago. 

"mouth pipette" is not a field that will be used when this query goes live for all SS2 submissions, right?

scripts/v6_queries/10x-query.json
malloryfreeberg: There is, in fact, an [Ontology Lookup Service (OLS)](http://ontology.dev.data.humancellatlas.org/index) :) Ontologized fields in the metadata schema are governed by different subsets of terms in the HCA ontology (HCAO), and those subsets are defined in the various `module/ontology/*.json` schemas. I am a novice user of the OLS, so @daniwelter might have a few tricks to offer for how to easily find what terms are valid for a given metadata field.

scripts/v6_queries/metadata_attachments.json
samanehsan: This file is used when we're creating subscriptions so that the notifications we receive will include these fields as "attachments". They won't affect the query at all!

scripts/v6_queries/smartseq2-query.json
samanehsan: I'm not sure, but @jishuxu or @dshiga might have a better idea here

scripts/v6_queries/metadata_attachments.json
samanehsan: To summarize the discussion that's been going on in slack, there's currently no way for us to query for the `sample_id` with a single expression here now that we need to find the sequencing protocol id from the protocol.json file and trace that back to the biomaterial source through the links.json. Going to remove this for now and make a ticket to add it in later.

scripts/v6_queries/10x-query.json
samanehsan: According to some discussion in slack, the `text` field is not affected by ontology validation, so we can use any term here. However according to @daniwelter: 
> it won’t be ontology mapped in that case (at least not until we have automatic annotation) - so it won’t be caught by semantic query expansion in search later on

samanehsan: I'll leave this as "analysis" for now but if there's a more appropriate term to use we can always update this later.

scripts/v6_queries/metadata_attachments.json
```diff
@@ -0,0 +1,14 @@
...
+    "type": "jmespath",
+    "expression": "files.project_json.content.project_core.project_shortname"
+  },
+  "submitter_id": {
+    "type": "jmespath",
+    "expression": "files.project_json.hca_ingest.submitter_id"
+  },
+  "sample_id": {
+    "type": "jmespath",
+    "expression": "files.biomaterial_json.biomaterials[?ends_with(content.describedBy, 'cell_suspension')].hca_ingest.document_id | [0]"
```
rexwangcc, line 12: per offline discussion, this might be changed in the future.

rexwangcc (approved):
LGTM!

PR #82, opened 2018-06-07
Set max_cromwell_retries in workflow inputs (790)

dshiga:
See https://elastc.com/c/dkauyyIF/790-make-ss2-have-cromwell-retries
and related PRs
https://github.com/HumanCellAtlas/skylab/pull/114
https://github.com/HumanCellAtlas/pipeline-tools/pull/51

- [x] Added or updated tests
- [ ] Updated documentation

### DIFF
--- a/kubernetes/listener-config.json.ctmpl
+++ b/kubernetes/listener-config.json.ctmpl
@@ -25,6 +25,7 @@
     {{ else }}
       "notification_token": "{{$listenerSecret.Data.notification_token}}",
     {{ end }}
+      "max_cromwell_retries": 0,
       "submit_wdl": "{{ env "PIPELINE_TOOLS_PREFIX" }}/adapter_pipelines/{{ env "SUBMIT_WDL_DIR" }}submit.wdl",
       "wdls": [
         {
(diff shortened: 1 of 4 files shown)


kubernetes/listener-config.json.ctmpl
```diff
@@ -25,6 +25,7 @@
     {{ else }}
       "notification_token": "{{$listenerSecret.Data.notification_token}}",
     {{ end }}
+      "max_cromwell_retries": 1,
```
rexwangcc: Should we parameterize this as well, or we only expect Cromwell retry at most for once?

rexwangcc (approved):
A minor question, otherwise LGTM!

kubernetes/listener-config.json.ctmpl
dshiga: Eventually, I'd like to make it a lot easier to render this whole file, given a list of parameter values. But since we are still manually rendering this, I don't think it's that helpful to make this a parameter. It's probably easier to just render the file and hand edit the retries value to be whatever you want.

rexwangcc: That's fair enough!

dshiga:
@rexwangcc, I changed max_cromwell_retries to zero in the config template so that mintegration tests will run without retries. The motivation for this is that we don't really want to rerun failed tasks in the mintegration test because most failures will not be transient and it will only slow down the test to repeat them. Let me know what you think.

rexwangcc:
It sounds good to me!

PR #83, opened 2018-06-08
Cw ultilize attachments from notification 784

rexwangcc:
Check here: https://elastc.com/c/ehji2BIV/784-utilize-new-information-received-in-notifications-from-blue

Note: I will make changes to the docstring style (Apply google style docstring) once the logic here is approved.
- [x] Added or updated tests
- [x] Updated documentation

---
Updates:
1. Based on several offline disuccsions with @dshiga and @samanehsan on this topic, since Cromwell v32(+) is more permissive on labels, i.e. it accepts all kind of valid strings in JSON with a maximum of 255 characters, we should remove the converter part from the `legalize_cromwell_labels` function in this PR. While backward compatibility is also important, since we are the only user of Lira at this time, and we are migrating to CaaS which has at least v33 Cromwell, we could go back to check how to improve it in the future(How can users use the latest version of Lira aginst an older version of Cromwell).
2. SmartSeq2 should only contain 1 `sample_id` in the attachments, so it makes sense to use the 1st element as the value of the `sample_id` label. For Optimus, it could be different, but we decided to consider this feature for Optimus later.


rexwangcc:
retest this please :octocat:

lira/lira_config.py
```diff
@@ -252,8 +253,8 @@ def __str__(self):
             self.MAX_CONTENT_LENGTH,
             self.wdls,
             self.version,
-            dss_url,
-            ingest_url)
+            self.dss_url,
```
rexwangcc, line 256: I'm not sure if this change is correct @samanehsan

samanehsan: I remember taking `dss_url` and `ingest_url` out of the config so I don't think they should be included here

lira/lira_utils.py
```diff
@@ -221,40 +240,47 @@ def legalize_cromwell_labels(label):
...
+    Args:
+        label (str | list): A string or a list of strings of key/value of labels need to be legalized.
+
+    Returns:
+        str: A converted, uglified but legal version of label.
 
-    :return str: A converted, uglified but legal version of label.
     """
+    if isinstance(label, list):
+        label = 'first-{}'.format(label[0])  # Only pick the first element in a list as a workaround here
```
samanehsan: If there is only one label in a list, I think we should not add the `first` prefix

lira/lira_config.py
rexwangcc: I'm not sure if it's necessary to have them here, are there special reasons about it? @dshiga

lira/lira_utils.py
rexwangcc: Sure, will do.

samanehsan: Also should this throw an error if there are multiple values for a label?

samanehsan:
For future work, I think it would be easier to separate formatting changes into their own PR so it's easier to tell what changes are related to the main ticket!

lira/test/notification.json
```diff
@@ -1,9 +1,74 @@
...
+      "submitter_id": "None",
+      "sample_id": [
+         "b1829a9d-6678-493b-bf98-01520f9bad52"
+      ],
+      "project_shortname": "Glioblastoma_medium_1000_cells"
+   },
+   "labels": {
+      "comment": "test"
+   }
+}
```
samanehsan: nit: please add a new line to the end of this file

lira/lira_utils.py
rexwangcc: This function is called atomically for both key and values for the labels, the only case it can get a list as input would be those sample_id in attachments: https://github.com/HumanCellAtlas/lira/blob/d875b496cd1b04a05f2f3d90d9a4d5e4d40bf69a/lira/test/notification.json#L66 

I was trying to make this as simple as possible since it was a very simple workaround. Since now we are using CaaS(v33+), and it is more permissive for values, do you think it's OK for use to remove this legalizer function?

lira/lira_config.py
samanehsan: From offline discussion, the `dss_url` and `ingest_url` are a part of the Lira config now so it makes sense to include them here since they are required fields

lira/lira_utils.py
```diff
@@ -239,20 +239,26 @@ def parse_github_resource_url(url):
...
+        ValueError: This will happen if the value of the label is a list, and the length of it is not equal to 1.
 
     """
+    cromwell_label_maximum_length = 255
+
     if isinstance(label, list):
-        label = 'first-{}'.format(label[0])  # Only pick the first element in a list as a workaround here
-    return label.lower().replace('_', '-').replace('.', '-').replace(':', '-')
+        if len(label) != 1:
+            raise ValueError('{} should contain exactly one element!'.format(label))
```
samanehsan: Can you include a test that tries to add a label where the value is a list containing more than one item?

lira/lira_utils.py
```diff
@@ -220,41 +239,54 @@ def parse_github_resource_url(url):
...
+        ValueError: This will happen if the value of the label is a list, and the length of it is not equal to 1.
 
-    :return str: A converted, uglified but legal version of label.
     """
-    return label.lower().replace('_', '-').replace('.', '-').replace(':', '-')
+    cromwell_label_maximum_length = 255
+
+    if isinstance(label, list):
+        if len(label) != 1:
+            raise ValueError('{} should contain exactly one element!'.format(label))
```
samanehsan, line 259: Can you include a test that tries to add a label where the value is a list containing more than one item?

samanehsan (approved):
I think this is good to merge after adding one more test case for labels where the value is a list: https://github.com/HumanCellAtlas/lira/pull/83/files#r195251000

PR #85, opened 2018-06-14
Fix a bug in lira_utils

rexwangcc:
Fix a bug that attachments field is None in notification body can break Lira.

- [x] Added or updated tests
- [ ] Updated documentation


samanehsan (approved):
Thanks @rexwangcc!

PR #86, opened 2018-06-26
Exclude results bundles in subscription query

samanehsan:
Update subscription queries to exclude results bundles. This will likely change since most of the fields that are currently in the analysis protocol file vary based on the pipeline run and should belong in an analysis process file.

- [ ] Added or updated tests
- [ ] Updated documentation


dshiga (approved):

samanehsan:
retest this please

PR #87, opened 2018-06-29
Update dependencies' version and refactor the version endpoint.

rexwangcc:
This PR:
1. Updates the version of the dependencies. (the latest version of gevent and connexion removes the misleading "monkey patch warning")
2. Updates the `/version` endpoint to reflect some recent changes to Lira's config, expose more useful information.

- [x] Added or updated tests
- [ ] Updated documentation


lira/test/test_get_version.py
```diff
@@ -28,17 +27,34 @@ def setUp(cls):
...
+
+    def test_get_version_can_fetch_correct_version_info(self):
+        response = self.client.get('/version')
+        json_response = json.loads(response.data.decode('utf-8'))
+        self.assertEqual(response.status_code, 200)
+
+        self.assertEqual(json_response.get('version_info').get('lira_version'), 'v0.1.0')
+        self.assertEqual(json_response.get('version_info').get('cromwell_tools_version'), cromwell_tools.__version__)
+        self.assertEqual(json_response.get('version_info').get('submit_wdl_version'), 'v0.1.5')
+        pass
```
samanehsan: Was the `pass` left over from debugging?

rexwangcc: Oops, that's for debugging, thanks for catching it!

samanehsan (approved):
I have one comment about a test, otherwise this looks good to me!

samanehsan:
Is the integration test failure because caas-dev was down earlier?

rexwangcc:
@samanehsan Yep, it was because the caas-dev was down. And it is still unavailable yet :(

samanehsan:
Ok, yeah I just noticed that as well :(

rexwangcc:
retest this please

start_lira.sh
```diff
@@ -13,6 +13,8 @@ fi
 #--workers: by default, it's set to 1, which is not enough, especially with sync workers, which means any of long
 #   requests will break/block the main process and cause CRITICAL TIMEOUTS problems.
 #   Now it's set to 2*AVAILABLE_CPU_CORES_OF_THE_CURRENT_MACHINE + 1, according to the suggestion of the gunicorn doc.
+#   The command will try to get the (2 * current number of CPU cores) + 1 first, if it failed, it will use 5 as the
```
samanehsan: It seems like you can remove "Now it's set to 2*AVAILABLE_CPU_CORES_OF_THE_CURRENT_MACHINE + 1, according to the suggestion of the gunicorn doc." from the README

start_lira.sh
```diff
@@ -13,6 +13,8 @@ fi
 #--workers: by default, it's set to 1, which is not enough, especially with sync workers, which means any of long
 #   requests will break/block the main process and cause CRITICAL TIMEOUTS problems.
 #   Now it's set to 2*AVAILABLE_CPU_CORES_OF_THE_CURRENT_MACHINE + 1, according to the suggestion of the gunicorn doc.
```
samanehsan: You can remove the line "Now it's set to 2*AVAILABLE_CPU_CORES_OF_THE_CURRENT_MACHINE + 1, according to the suggestion of the gunicorn doc." from the README

start_lira.sh
samanehsan: Did we previously run into an issue where setting the number of workers to 2*AVAILABLE_CPU_CORES_OF_THE_CURRENT_MACHINE + 1 failed? I'm just trying to better understand the reason behind this change.

rexwangcc: We haven't run into any issues with the command `$((2 * $(getconf _NPROCESSORS_ONLN) + 1))`. The reason behind this change is mainly for a better portability:
`getconf _NPROCESSORS_ONLN` is only available in Linux OS/Mac OS, for BSD OS we have to use `getconf NPROCESSORS_ONLN`. If both of them fail, (such as windows), we have to give it a default value, otherwise, Lira will fail to start.

samanehsan: Thanks for the explanation!

samanehsan (approved):

rexwangcc:
The Mintegration test failed:
```
  File "/usr/local/lib/python2.7/site-packages/hca/upload/upload_area_urn.py", line 12, in __init__
    assert urnbits[0:3] == ['dcp', 'upl', 'aws'], "URN does not start with 'dcp:upl:aws': %s" % (urn,)
AssertionError: URN does not start with 'dcp:upl:aws': s3://org-humancellatlas-upload-integration/xxxxx/
```
It was beacuse the master branch of pipeline-tools is still using an older version of hca cli, this should now be fixed by the merge of https://github.com/HumanCellAtlas/pipeline-tools/pull/56

rexwangcc:
ret

PR #88, opened 2018-07-05
Increase lira deployment replicas

samanehsan:
Increase the number lira deployment replicas from 1 to 3.


kubernetes/listener-deployment.yaml
```diff
@@ -3,7 +3,7 @@ kind: Deployment
 metadata:
   name: listener
 spec:
-  replicas: 1
+  replicas: 3
```
dshiga: Now that we have the ctmpl file, I wonder if there is any reason to keep this yaml file. I don't think we actually use it for anything, do we?

samanehsan: That's a good point! I'll go ahead and remove it

dshiga (approved):
Can you rerun the mintegration test? I'm assuming it failed because caas was down.

samanehsan:
retest this please

samanehsan:
retest this please

samanehsan:
retest this please

PR #93, opened 2018-08-08
Update smartseq2 query

samanehsan:
Update smartseq2 query based on changes to the structure of the bundle manifest file that move files out of files.json, process.json, etc. and into the top-level of the bundle manifest.

Closes https://github.com/HumanCellAtlas/secondary-analysis/issues/104

### DIFF
--- a/scripts/v7_queries/smartseq2-query.json
+++ b/scripts/v7_queries/smartseq2-query.json
@@ -4,49 +4,49 @@
       "must": [
         {
           "match": {
-            "files.library_preparation_protocol_0.library_construction_approach.text": "Smart-seq2"
+            "files.library_preparation_protocol_json.library_construction_approach.text": "Smart-seq2"
 
           }
         },
         {
           "match": {
-            "files.sequencing_protocol_0.paired_ends": true
+            "files.sequencing_protocol_json.paired_ends": true
           }
         },
         {
           "match": {
-            "files.donor_organism_0.biomaterial_core.ncbi_taxon_id": 9606
+            "files.donor_organism_json.biomaterial_core.ncbi_taxon_id": 9606
           }
         }
       ],
       "should": [
         {
           "match": {
-            "files.dissociation_protocol_0.dissociation_method.text": "fluorescence-activated cell sorting"
+            "files.dissociation_protocol_json.dissociation_method.text": "fluorescence-activated cell sorting"
           }
         },
         {
           "match": {
-            "files.dissociation_protocol_0.dissociation_method.text": "mouth pipette"
+            "files.dissociation_protocol_json.dissociation_method.text": "mouth pipette"
           }
         }
       ],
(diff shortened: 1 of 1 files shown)


scripts/v7_queries/smartseq2-query.json
```diff
@@ -0,0 +1,57 @@
...
+        {
+          "match": {
+            "files.dissociation_protocol_0.dissociation_method.text": "mouth pipette"
+          }
+        }
+      ],
+      "must_not": [
+        {
+          "match": {
+            "files.analysis_protocol_0.protocol_type.text": "analysis"
```
samanehsan: Since there is no analysis json file in the example bundle (which is expected since it hasn't gone through secondary-analysis yet), I'm not sure what this query string should be. Does this look correct @malloryfreeberg?

malloryfreeberg: The path to the field looks as correct to me as I would guess (`files.analysis_protocol_0.protocol_type.text`), but I don't see the term "analysis" under the ontology curie "EFO:0002694", which is the term governing this field. @daniwelter can maybe help either identify what text/curie we should use here or add "analysis" to the ontology.

samanehsan: Thanks @malloryfreeberg! Last time we agreed that the term `analysis` was okay to use as a text field because there was no equivalent ontology term. But perhaps we should add one so that we don't have to rely on matching this string.

scripts/v7_queries/smartseq2-query.json
```diff
@@ -0,0 +1,57 @@
+{
+  "query": {
+    "bool": {
+      "must": [
+        {
+          "match": {
+            "files.library_preparation_protocol_json.library_construction_approach.text": "Smart-seq2"
```
rexwangcc, line 7: Should there be a `.content.` before `library_construction_approach` and after `library_preparation_protocol_json`?

rexwangcc: And this `content` question for all of the query strings too.

rexwangcc (approved):
I have a question about `.content.`, but since I was able to use this query to fetch the latest primary bundles in dev dss, it LGTM! 

By the way, I cannot find any bundles with this query in integration DSS, I need to make sure that is expected(no new bundles in integration DSS), otherwise this is a false negative.

scripts/v7_queries/smartseq2-query.json
samanehsan: The new bundle structure does not include a `content` key in these files

rexwangcc: Cool, that makes sense then, thanks for the explanation here

samanehsan:
Some new bundles were added to the integration DSS recently and this query matches them!

PR #94, opened 2018-08-22
Add cromwell_url as a input parameter to adapter workflows.

rexwangcc:
This PR adds `cromwell_url` as an input parameter to be passed into the adapter workflows that Lira starts.

- [X] Added or updated tests
- [ ] Updated documentation

### DIFF
--- a/lira/lira_utils.py
+++ b/lira/lira_utils.py
@@ -172,7 +172,8 @@ def compose_inputs(workflow_name, uuid, version, lira_config):
         workflow_name + '.submit_url': lira_config.ingest_url,
         workflow_name + '.schema_url': lira_config.schema_url,
         workflow_name + '.use_caas': lira_config.use_caas,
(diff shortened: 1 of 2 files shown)


samanehsan (approved):
Looks good to me @rexwangcc!

rexwangcc:
As discussed with @samanehsan I'm going to merge this PR with the failed Mintegration test result. 

The Mintegration test is failing at `create_submission` step, because the bundle-structure change is backward incompatible and Ingest now only recognizes the new bundle structure:
>  during that request, the process is being linked to the bundle manifest and to the file metadata listed in that bundle manifest. Before there is only one file for all the files metadata. And we're using the uuids of those file metadata for their corresponding data files in data store. We changed that, and along that change, it's retrieving the uuids from the bundle manifest differently, it's possible that it couldn't find the uuid of the file metadata and couldn't link it to the process as input file.

PR #98, opened 2018-08-28
Add updated 10x query

dshiga:
- [ ] Added or updated tests
- [ ] Updated documentation

See https://app.zenhub.com/workspace/o/humancellatlas/secondary-analysis/issues/209

### DIFF
--- a/scripts/vx_queries/10x-query.json
+++ b/scripts/vx_queries/10x-query.json
@@ -1,0 +1,41 @@
+{
+  "query": {
+    "bool": {
+      "must": [
+        {
+          "match": {
+            "files.library_preparation_protocol_json.library_construction_approach.ontology": "EFO:0009310"
+
+          }
+        },
+        {
+          "match": {
+            "files.donor_organism_json.biomaterial_core.ncbi_taxon_id": 9606
+          }
+        }
+      ],
+      "must_not": [
+        {
+          "match": {
+            "files.process_json.process_type.text": "analysis"
+          }
+        },
+        {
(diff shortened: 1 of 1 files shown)


scripts/vx_queries/10x-query.json
```diff
@@ -0,0 +1,41 @@
+{
+  "query": {
+    "bool": {
+      "must": [
+        {
+          "match": {
+            "files.library_preparation_protocol_json.library_construction_approach.ontology": "EFO:0009310"
```
rexwangcc, line 7: Based on https://www.ebi.ac.uk/ols/ontologies/efo/terms?iri=http%3A%2F%2Fwww.ebi.ac.uk%2Fefo%2FEFO_0009310 and
```json
"library_construction_approach": {
                "ontology": "EFO:0009310",
                "ontology_label": "10X v2 sequencing",
                "text": "10x v2"
              }
```
I think this should be the right way to query 10x bundle. My curiouse question here is why we don't use:
`"files.library_preparation_protocol_json.library_construction_approach.text": "10x v2"` (like what we did in ss2 query), is it because the text is not as reliable as the ontology?

rexwangcc (approved):
I left a curious question, otherwise LGTM!

scripts/vx_queries/10x-query.json
dshiga: It would probably be okay either way, but this way protects us a bit in case anyone decides to change the text representation, e.g. to "10x_v2" or something.

PR #100, opened 2018-09-04
Fix the subscription queries so that result bundles won't trigger notifications.

rexwangcc:
- [ ] Added or updated tests
- [ ] Updated documentation


dshiga (approved):
Did you test the query against blue box to make sure it excludes analysis bundles?

samanehsan (approved):

rexwangcc:
@dshiga Yep, I tested both queries, they excluded the analysis bundles and also included the primary bundles.

PR #107, opened 2018-09-26
Update to the new DCP auth method to work with subscriptions.

rexwangcc:
- [ ] Added or updated tests
- [x] Updated documentation


scripts/subscription-create.sh
```diff
@@ -1,53 +1,28 @@
...
-fi
-if [ -z $listener_secret ]; then
-  echo "listener_secret is required"
-  error=1
-fi
-if [ -z $query_json ]; then
-  echo "query_json is required"
-  error=1
-fi
+config_script=$1
```
samanehsan: This script is used in some of the Lira deployment set-up scripts, e.g: https://github.com/HumanCellAtlas/lira/blob/master/kubernetes/create_subscriptions.sh#L80

rexwangcc: Ah, didn't aware of that!

rexwangcc: I'm going to change there too!

scripts/requirements.txt
```diff
@@ -1,2 +1,3 @@
-httplib2
-oauth2client
+cryptography
```
samanehsan: Let's include versions for these dependencies

scripts/subscribe.py
```diff
@@ -0,0 +1,273 @@
...
+
+    if command == 'create':
+        create_subscription(args)
+    elif command == 'list':
+        list_subscription(args)
+    elif command == 'delete':
+        delete_subscription(args)
+
+
+class DCPAuthClient(object):
```
samanehsan: Can we move this class into a separate file? I think that separation would make it easier to read the code in this script.

scripts/config-template.sh
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 
 dss_url=test
-green_url=test
+lira_url=test
```
samanehsan: Can we replace the example values for `dss_url` and `lira_url` here with example urls? I can never remember whether these urls are for a specific endpoint or if they're just the base urls for these services.

kubernetes/create_subscriptions.sh
```diff
@@ -77,9 +77,27 @@ SMART_SEQ_2_QUERY=${SMART_SEQ_2_QUERY:-"v6_queries/smartseq2-query.json"}
...
 #echo "Creating ss2 subscription"
-SS2_SUBSCRIPTION_ID=$(bash subscription-create.sh ${DSS_URL} ${GREEN_URL} ${BLUEBOX_KEY_PATH} ${LIRA_SECRET} ${SMART_SEQ_2_QUERY} ${HMAC_KEY_FILE} ${ADDITIONAL_METADATA} ) # | jq '.[]')
+SS2_SUBSCRIPTION_ID=$(python3 subscribe.py create --dss_url="${DSS_URL}" \
+                            --key_file="${BLUEBOX_KEY_PATH}" \
+                            --google_project="${GCLOUD_PROJECT}" \
+                            --replica="gcp" \
+                            --callback_base_url="${LIRA_URL}" \
+                            --query_json="${SMART_SEQ_2_QUERY}" \
+                            --hmac_key_id="${HMAC_KEY_FILE}" \
+                            --hmac_key="${LIRA_SECRET}" \
```
samanehsan: I'm not sure that this is the correct value to use for the `hmac_key`: https://github.com/HumanCellAtlas/lira/blob/d239a77f684b8fcc5834bea76d526165557e2015/kubernetes/create_subscriptions.sh#L64

scripts/subscriptions-list.sh
```diff
@@ -1,4 +1,11 @@
 #!/usr/bin/env bash
 
-source config.sh
-./subscriptions-list.py "$dss_url?replica=gcp" $key_file
\ No newline at end of file
+# Usage: bash subscriptions-list.sh config.sh
+
+config_script=$1
```
samanehsan: `config_script` makes me think this parameter is asking for a script...how about calling this `script_config` instead?

scripts/subscribe.py
```diff
@@ -0,0 +1,273 @@
...
+    del args['command']
+
+    # Make sure we are using the /subscriptions endpoint of DSS
+    args['dss_url'] = '{url}/subscriptions'.format(url=args['dss_url'].strip('/'))
+
+    return command, args
+
+
+if __name__ == '__main__':
+    main(sys.argv[1:])
```
samanehsan: Why do we discard the first argument here? Can you add a comment to clarify?

scripts/subscribe.py
```diff
@@ -0,0 +1,273 @@
...
+    print(response.json())
+
+
+def parser(arguments):
+    main_parser = argparse.ArgumentParser()
+
+    subparsers = main_parser.add_subparsers(help='All available commands', dest='command')
+    create = subparsers.add_parser('create', help='Create a new subscription in DSS.')
+    delete = subparsers.add_parser('delete', help='Delete an existing subscription in DSS.')
+    list = subparsers.add_parser('list', help='List all subscriptions in DSS.')
```
samanehsan: `list` is a reserved keyword in python...how about renaming this variable to `get` or `list_subscriptions`?

scripts/subscribe.py
```diff
@@ -0,0 +1,273 @@
...
+    for sub_command in subscribe_sub_commands:
+        sub_command.add_argument('--dss_url',
+                                 help='URL to the HCA DCP Data Storage System API.',
+                                 required=True)
+        sub_command.add_argument('--key_file',
+                                 help='JSON file containing Storage Service credentials.',
+                                 required=True)
+        sub_command.add_argument('--replica',
+                                 help='Which replica to work on, use "gcp" by default. ["gcp", "aws"]',
+                                 # required=True,
```
samanehsan: Can we remove this comment?

scripts/subscribe.py
```diff
@@ -0,0 +1,273 @@
...
+
+    # Add specific argument for `create` command
+    create.add_argument('--callback_base_url',
+                        help='Lira endpoint for receiving notifications.',
+                        required=True)
+    create.add_argument('--query_json',
+                        help='JSON file containing the ElasticSearch Subscription query to register.',
+                        required=True)
+
+    HMAC_auth_group = create.add_mutually_exclusive_group(required=True)
```
samanehsan, line 148: Just for my own understanding, what does `add_mutually_exclusive_group` do? Does that mean you can only provide one of the arguments in the group?

scripts/subscriptions-list.sh
rexwangcc: Hmmm, this parameter is asking for a script indeed (a `.sh` file with a lot of values defined within it), but `script_config` does make more sense here! will do

scripts/subscribe.py
```diff
@@ -0,0 +1,273 @@
...
+    if command == 'create':
+        create_subscription(args)
+    elif command == 'list':
+        list_subscription(args)
+    elif command == 'delete':
+        delete_subscription(args)
+
+
+class DCPAuthClient(object):
+    audience = "https://dev.data.humancellatlas.org/"
```
samanehsan: Does this belong in the `__init__` function? Also this would be more flexible if this value was configurable as well

scripts/subscribe.py
rexwangcc: Will add a comment to clarify, basically it is:
```
import sys
print(sys.argv, len(sys.argv))

> python print_args.py
['print_args.py'] 1

> python print_args.py foo and bar
['print_args.py', 'foo', 'and', 'bar'] 4

> python print_args.py "foo and bar"
['print_args.py', 'foo and bar'] 2

> python print_args.py "foo and bar" and baz
['print_args.py', 'foo and bar', 'and', 'baz'] 4
```
so `sys.argv[1:]` excluded " print_args.py" if you do `python print_args.py a b c`

scripts/subscribe.py
rexwangcc: Yes, it will make the arguments within the group mutually exclusive at the same time, according to https://docs.python.org/3/library/argparse.html#mutual-exclusion

scripts/subscribe.py
rexwangcc: Forgot to remove it, will do!

scripts/subscribe.py
rexwangcc: `get` makes more sense, will do!

scripts/subscribe.py
rexwangcc: Will do!

scripts/config-template.sh
rexwangcc: Good call, will improve!

kubernetes/create_subscriptions.sh
rexwangcc: Good catch, that is the value of the notification token we had. Unfortunately, since I copied this over, does it mean the script was wrong along the time? :(

scripts/subscribe.py
rexwangcc: Based on my conversation with @Bento007 this should be hard-coded as it is, except "https://data.humancellatlas.org/" for prod? You are right, even then we can make the script smarter to figure out which value to use!

scripts/subscribe.py
```diff
@@ -0,0 +1,273 @@
...
+                   }
+        additional_headers = {'kid': service_credentials["private_key_id"]}
+        signed_jwt = jwt.encode(payload, service_credentials["private_key"], headers=additional_headers,
+                                algorithm='RS256').decode()
+        return signed_jwt
+
+    def verify_jwt(self, token, audience, trusted_google_project):
+        assert token
+        try:
+            openid_provider = "humancellatlas.auth0.com"
```
samanehsan: Instead of having the `openid_provider` hard-coded here, we can pass it in as a parameter

kubernetes/create_subscriptions.sh
samanehsan: That's what I was thinking...but somehow our subscriptions were still working 🤔

scripts/subscribe.py
rexwangcc: I think this won't change within the context of HCA, so I prefer to leave it hard-coded for now?

kubernetes/create_subscriptions.sh
rexwangcc: Just confirmed it was broken... The HMAC_KEY_FILE's content is like:
```
{
  "request_id": "xxxxx",
  "lease_id": "",
  "lease_duration": 2764800,
  "renewable": false,
  "data": {
    "XXX": "YYY"
  },
  "warnings": null
}
```
It cannot work with the script. I think our subscription is working just simply because we have never used this set-up script to make subscriptions anywhere(not in Mintegration test, nor in Jenkins jobs), we have been manually creating the subscriptions.

Fixing the script requires using `jq` or alternatively some refactoring of the script, I will try to do it now

rexwangcc: I fixed the script (partially). To make the script support both auth_token and HMAC would be a very tricky problem in Bash, so I dropped the support for auth_token in the script and added a comment there. What do you think?

rexwangcc:
@samanehsan I made the changes based on your comments, let me know what do you think!

scripts/requirements.txt
```diff
@@ -1,2 +1,3 @@
-httplib2
-oauth2client
+cryptography====2.3.1
```
samanehsan: This should be be `==` here

rexwangcc: 🤦‍♂️ Will fix...

kubernetes/create_subscriptions.sh
```diff
@@ -61,6 +61,7 @@ docker run -it --rm \
                @/working/"${BLUEBOX_SUBSCRIPTION_KEY}"
 
 # Get the lira secret from vault
+# Note: This value is not used in this script for now. This script only supports making subscriptions with HMAC method
```
samanehsan, line 64: @rexwangcc can you please create a ticket so that we remember to come back and update this script?

rexwangcc: Sure! Making it now!

scripts/subscribe.py
samanehsan: @rexwangcc just wondering why does this include `dev` and not `integration` and `staging` as well?

scripts/subscribe.py
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env python
+import sys
+
+import argparse
+import json
+import requests
+
+try:
+    from DCPAuthClient import DCPAuthClient
+except ModuleNotFoundError:
```
samanehsan, line 10: I feel like this try/catch is unnecessary here because this script would throw an import error otherwise, but that's just my preference for less code 🙂

samanehsan (approved):
Looks good to me @rexwangcc! :octocat:

scripts/subscribe.py
rexwangcc: Yeah it is unnecessary from the functionalities perspective. The only purpose for it was that I wanted to re-raise the error with a more user-friendly message(while shadowing the actual error msg may not be a good idea in some cases), e.g.:
```
Traceback (most recent call last):
  File "subscribe.py", line 8, in <module>
    from DCPAuthClient import DCPAuthClient
ModuleNotFoundError: No module named 'DCPAuthClient'
```
v.s.
```
Cannot find the DCPAuthClient within this folder!
```

But we can defo remove it.

rexwangcc:
Since DSS has fixed the issue with their `/GET` endpoint, I'm going to merge this PR!

PR #114, opened 2018-11-27
Ra first gitlab job

rhiananthony:
- [ ] Added or updated tests
- [ ] Updated documentation


.gitignore
```diff
@@ -12,3 +12,5 @@ firecloud-tools
 kubernetes/vault-edit
 scripts/hmac_keys
 .DS_Store
+scripts/bluebox-subscription-manager-key-staging.json
```
samanehsan: Let's add the subscription manager keys for the other environments to the `.gitignore` as well

deploy/scripts/certbot-route53.sh
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env bash
+
+# This script was taken from this repo: https://github.com/jed/certbot-route53.git
+# It was copied locally is built on the fly because no docker image exists with this code
+# I copied the file locally because it is self contained and there is no reason to grab the whole repo
+# every time that we want to do a deployment of a new cert
```
samanehsan: Since this script is now slightly modified from the original, I think it would be useful to briefly mention what was changed and why.

deploy/config_files/config.sh.ctmpl
```diff
@@ -0,0 +1,184 @@
...
+
+    export GCLOUD_PROJECT="broad-dsde-mint-${LIRA_ENVIRONMENT}" # all valid envs - broad-dsde-mint-dev, broad-dsde-mint-test, broad-dsde-mint-integration, broad-dsde-mint-staging, hca-dcp-pipelines-prod
+
+    export DSS_URL="https://dss.integration.data.humancellatlas.org/v1"
+    export SCHEMA_URL="http://schema.integration.data.humancellatlas.org/"
+    export INGEST_URL="http://api.ingest.integration.data.humancellatlas.org/"
+    export DOMAIN="pipelines.integration.data.humancellatlas.org"
+{{end}}
+
+{{ if eq $ENVIRONMENT "test"}}
```
samanehsan: Could we change this to check if the environment is "test" or "integration"? I was thinking the "integration" deployment would eventually replace "test" so their configuration would be the same.

deploy/gitlab/deploy_lira.sh
```diff
@@ -0,0 +1,93 @@
...
+export VAULT_WRITE_TOKEN_PATH="/gitlab-runner/.vault-write-token"
+export VAULT_TOKEN="$(cat ${VAULT_READ_TOKEN_PATH})"
+
+export WORK_DIR=$(pwd)
+export CONFIG_DIR=${WORK_DIR}/deploy/config_files
+export DEPLOY_DIR=${WORK_DIR}/deploy/gitlab
+export SCRIPTS_DIR=${WORK_DIR}/deploy/scripts
+
+echo "Rendering deployment configuration file"
+sh /usr/local/bin/render-ctmpls.sh -k ${CONFIG_DIR}/config.sh.ctmpl
```
samanehsan: Where does the `LIRA_ENVIRONMENT` variable used in `config.sh.ctmpl` get set? Is it in gitlab?

deploy/config_files/config.sh.ctmpl
```diff
@@ -0,0 +1,184 @@
+#!/usr/bin/env bash
+{{with $ENVIRONMENT := env "LIRA_ENVIRONMENT"}}
+{{ if eq $ENVIRONMENT "dev" }}
+    export LIRA_VERSION="v0.14.1"
```
samanehsan, line 4: I've seen scripts import variables from a config.sh file, but the variables were not using the `export` keyword. Is it optional? E.g. this [script](https://github.com/HumanCellAtlas/lira/blob/master/scripts/subscription-create.sh#L9) is using these variables: https://github.com/HumanCellAtlas/lira/blob/master/scripts/config-template.sh

deploy/gitlab/deploy_lira.sh
samanehsan: Nevermind! I see it is defined in `.gitlab-ci.yml`

.gitlab-ci.yml
```diff
@@ -1,6 +1,12 @@
...
+  - deployment
 
-dev:
-  stage: dev
+deploy_lira:
+  tags:
+    - green
+  stage: deployment
   script:
+    - export LIRA_ENVIRONMENT=staging
```
samanehsan, line 31: How will the deployments work with other environments?

deploy/gitlab/deploy_lira.sh
```diff
@@ -0,0 +1,93 @@
...
+
+# Import the variables from the config files
+source ${CONFIG_DIR}/config.sh
+
+echo "Retrieving caas service account key"
+vault read -format=json "${CAAS_KEY_PATH}" | jq .data > "${CONFIG_DIR}/${CAAS_KEY_FILE}"
+
+echo "Authenticating with the service account"
+gcloud auth activate-service-account --key-file "${CONFIG_DIR}/${CAAS_KEY_FILE}"
```
samanehsan, line 24: I know this is existing code, but can you explain why this script needs to authenticate with the CAAS service account in order to deploy Lira?

deploy/gitlab/deploy_lira.sh
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env bash
+
+export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/google-cloud-sdk/bin
+
+export VAULT_READ_TOKEN_PATH="/gitlab-runner/.vault-read-token"
```
samanehsan, line 5: Why is there a separate vault read token? Is it to prevent accidentally overwriting values in vault?

deploy/gitlab/deploy_lira.sh
```diff
@@ -0,0 +1,93 @@
...
+    kubectl create secret generic ${LIRA_CONFIG_SECRET_NAME} \
+            --from-file=config="${CONFIG_DIR}/${LIRA_CONFIG_FILE}" \
+            --namespace "${KUBERNETES_NAMESPACE}"
+fi
+
+echo "Generating Lira deployment file"
+sh /usr/local/bin/render-ctmpls.sh -k "${CONFIG_DIR}/${LIRA_DEPLOYMENT_YAML}.ctmpl"
+
+echo "Deploying Lira"
+kubectl apply -f ${CONFIG_DIR}/lira-deployment.yaml \
```
samanehsan: I think this is existing code, but `lira-deployment.yaml` should be `${LIRA_DEPLOYMENT_YAML}`

deploy/jenkins/generate_certs.sh
```diff
@@ -1,100 +1,79 @@
...
 docker build -t certbot .
 
+echo "Executing the certbot script to create a cert"
 docker run \
     -e AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID} \
     -e AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY} \
-    -v "${PWD}/certs":/working \
-    -v "${PWD}/certbot-route53.sh":/working/certbot-route53.sh \
+    -v $(pwd)/certs:/certs \
+    -v $(pwd)/certbot-route53.sh:/certs/certbot-route53.sh \
```
samanehsan: Does this volume mount work? The `certbot-route53.sh` script is no longer in the same directory as this script

samanehsan:
Can you update this [README.md](https://github.com/HumanCellAtlas/lira/blob/ra_first_gitlab_job/kubernetes/README.md) and add it to the lira/deploy directory?

samanehsan:
I think we can move these two scripts into the lira/scripts directory and then just remove the "kubernetes" directory: https://github.com/HumanCellAtlas/lira/tree/ra_first_gitlab_job/kubernetes

We may want to organize the lira/scripts directory in a follow-up PR since there are scripts for setting up Cromwell-as-a-Service and scripts for creating data store subscriptions that have nothing to do with each other.

samanehsan:
Can you clarify why the scripts to create/update the Lira service, ingress and deployment are now in one script? I thought they were originally separated because most of the time we only need to update the deployment.

.gitignore
rhiananthony: Done :)

deploy/scripts/certbot-route53.sh
rhiananthony: Done :)

deploy/config_files/config.sh.ctmpl
rhiananthony: Good idea. Done!

deploy/config_files/config.sh.ctmpl
rhiananthony: Using "export" makes the variable available to sub-processes.

The variable name is available to any process you run from that shell process. If you want a process to make use of this variable, use export, and run the process from that shell.

Without export, the variable scope is restricted to the shell, and is not available to any other process. You would use this for (say) loop variables, temporary variables etc.

It's important to note that exporting a variable doesn't make it available to parent processes.

.gitlab-ci.yml
rhiananthony: We will have to modify the gitlab.yml file

deploy/gitlab/deploy_lira.sh
rhiananthony: This was a check in the process of the deployment. If you can't authenticate, then there is no reason to continue the deployment.

deploy/gitlab/deploy_lira.sh
rhiananthony: Yes. It was consistent with the principle of least access. Either one token would have a more complex vault profile which would allow it to read the entire mint directory of vault and write to only the location of the TLS cert keys OR we have 2 vault tokens, one which reads and one which can write. In this case it was the decision that Henry made.

deploy/gitlab/deploy_lira.sh
rhiananthony: Yes you are right :) Great catch!

.gitlab-ci.yml
rhiananthony: We should also have a follow up ticket as to how we may be able to improve this in the future. We may be able to have a true pipeline of deployments with a stage for each environment.

deploy/config_files/config.sh.ctmpl
```diff
@@ -0,0 +1,170 @@
...
+    export DSS_URL="https://dss.integration.data.humancellatlas.org/v1"
+    export SCHEMA_URL="http://schema.integration.data.humancellatlas.org/"
+    export INGEST_URL="http://api.ingest.integration.data.humancellatlas.org/"
+    export DOMAIN="pipelines.integration.data.humancellatlas.org"
+{{end}}
+
+{{ if eq $ENVIRONMENT "staging" }}
+    export LIRA_VERSION="v0.14.1"
+
+    export PIPELINE_TOOLS_VERSION="v0.42.0"
```
samanehsan: This version is now `v0.42.1`

.gitlab-ci.yml
```diff
@@ -1,6 +1,41 @@
...
+  stage: staging
+  script:
+    - export LIRA_ENVIRONMENT=staging
+    - bash deploy/gitlab/deploy_lira.sh
+  when: manual
+
+deploy_lira:
+  tags:
+    - green
+  stage: prod
```
pshapiro4broad, line 38: do you have to add `prod` to the `stages` section above?

start_lira.sh
```diff
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 
-port=$1
+port=${1}
```
pshapiro4broad: the `{` `}` are unnecessary here and below around `port`

deploy/jenkins/generate_certs.sh
rhiananthony: Great catch! This would not work for jenkins. Fixed :)

deploy/config_files/config.sh.ctmpl
rhiananthony: Updated. Actually to v0.43.0

.gitlab-ci.yml
rhiananthony: It's been added

rhiananthony:
The primary reason that the three scripts which were used in the Jenkins job are now united into one script is because gitlab makes it difficult to have jobs separated in the same way that Jenkins does. Because of the different workflow, the jobs made more sense as one than three. Also, if the components don't change then there is no harm in running the full script every time. It may take a few extra seconds, but overall should not detract too much from the deployment.

start_lira.sh
rhiananthony: Fixed

documentation/README.md
```diff
@@ -1,28 +1,27 @@
...
-## Upgrading
-### Lira Deployments (deploy_lira.sh)
-This script does the following:
-1. Retrieving caas service account key from vault
-2. Renders the Lira config file
-3. Deploys the lira config file with the caas key to the Kubernetes cluster
-4. Generates the lira deployment file
-5. Deploys the lira-deployment.yaml file to the Kubernetes cluster
+## Installing or Upgrading
+Generally the only file which will need to be modified when updating a deployment will be the
```
samanehsan: Thanks for updating the README! I think it makes sense to put this in the `deploy` directory since it's referring to scripts in that directory

deploy/config_files/config.sh.ctmpl
```diff
@@ -0,0 +1,170 @@
...
+{{end}}
+
+{{if (or (eq $ENVIRONMENT "test") (eq $ENVIRONMENT "integration")) }}
+    export LIRA_VERSION="v0.14.1"
+
+    export PIPELINE_TOOLS_VERSION="4k-scaling-test-2018-11-06-with-papiv2"
+
+    export NUMBER_OF_REPLICAS="3"
+
+    export GENERATE_CERTS="true"
```
samanehsan, line 35: We'll need to remember to change this back to "false" after regenerating the cert!

samanehsan (approved):
This looks great @rhiananthony! I just have one minor comment about moving the deployment README into the "deploy" directory: https://github.com/HumanCellAtlas/lira/pull/114#discussion_r240778673 and then I think this is good to merge :octocat:

deploy/config_files/config.sh.ctmpl
rhiananthony: 👍

documentation/README.md
rhiananthony: Done :)

samanehsan:
I just realized that because this PR changes the location of the `ctmpl` files and removes the `kubernetes` directory, all of the old file paths in the integration test need to be updated in order fot the test to pass, e.g: https://github.com/HumanCellAtlas/secondary-analysis/blob/master/tests/integration_test/integration_test.sh#L484

PR #133, opened 2019-02-13
Make query for 10xv2 data more specific

samanehsan:
### Purpose
Make the query for 10xv2 data more specific so that secondary analysis only receives notifications for data that can be processed by the Cellranger and Optimus pipelines.

---
### Changes
Update the query to get only single cell data with a 3 prime tag end bias and exclude anything from CITEseq assays.

---
### Review Instructions
Should there be a single `sequencing_approach.ontology` that the query matches on, or should the query just exclude CITEseq data?

The 10xv2 datasets in production have various values for the `sequencing_approach.ontology` field. For example, the Peer dataset is using `EFO:0008995 (10x sequencing)` and the fetal/maternal interface dataset is using `EFO:0008440 (Tag based single cell RNA sequencing)`. Both have been analyzed by the Cellranger pipeline.

---
### PR Checklist

---
### Follow-up Discussions
- No follow-up discussions.

### DIFF
--- a/scripts/vx_queries/10x-query.json
+++ b/scripts/vx_queries/10x-query.json
@@ -4,7 +4,7 @@
       "must": [
         {
           "match": {
-            "files.library_preparation_protocol_json.library_construction_approach.ontology": "EFO:0009310"
+            "files.library_preparation_protocol_json.library_construction_approach.ontology_label": "10X v2 sequencing"
           }
         },
         {
@@ -26,7 +26,7 @@
       "must_not": [
         {
           "match": {
-            "files.sequencing_protocol_json.sequencing_approach.ontology": "EFO:0009294"
+            "files.sequencing_protocol_json.sequencing_approach.ontology_label": "CITE-seq"
           }
         },
         {
--- a/scripts/vx_queries/10x-query.json
+++ b/scripts/vx_queries/10x-query.json
@@ -5,7 +5,16 @@
         {
           "match": {
             "files.library_preparation_protocol_json.library_construction_approach.ontology": "EFO:0009310"
-
+          }
+        },
+        {
+          "match": {
+            "files.library_preparation_protocol_json.end_bias": "3 prime tag"
+          }
+        },
+        {
+          "match": {
+            "files.library_preparation_protocol_json.nucleic_acid_source": "single cell"
           }
         },
         {
@@ -15,6 +24,11 @@
         }
       ],
       "must_not": [
+        {
+          "match": {
+            "files.sequencing_protocol_json.sequencing_approach.ontology": "EFO:0009294"
+          }
+        },
         {
           "match": {
             "files.analysis_process_json.process_type.text": "analysis"
@@ -38,4 +52,3 @@
     }
   }
 }
-


scripts/vx_queries/10x-query.json
```diff
@@ -15,6 +24,11 @@
         }
       ],
       "must_not": [
+        {
+          "match": {
+            "files.sequencing_protocol_json.sequencing_approach.ontology": "EFO:0009294"
```
ambrosejcarr: Is there any way to document that this is CITE-seq?

ambrosejcarr (approved):

scripts/vx_queries/10x-query.json
```diff
@@ -5,7 +5,16 @@
         {
           "match": {
             "files.library_preparation_protocol_json.library_construction_approach.ontology": "EFO:0009310"
-
+          }
+        },
+        {
+          "match": {
+            "files.library_preparation_protocol_json.end_bias": "3 prime tag"
```
malloryfreeberg, line 12: Another valid value for this field is "3 prime end bias". I'm not entirely sure how this is different than "3 prime tag", but might contributors put this value in for 3' 10x v2?

scripts/vx_queries/10x-query.json
daniwelter: I believe consensus among wranglers is to tag sequencing_approach with either 'tag-based single cell sequencing' or 'full length single cell sequencing' so I'm not sure whether this exclusion would necessarily work. @malloryfreeberg could you comment?

daniwelter (approved):

scripts/vx_queries/10x-query.json
malloryfreeberg: I'm not sure what we should call this in the future. For now, the one dataset we have of this is referred to as CITEseq (EFO:0009294) in the metadata. So, the query is sufficient based on current dataset, but might need to be updated (again!) in the future if we figure out a way to standardize representing cell hashing experiments.

scripts/vx_queries/10x-query.json
ambrosejcarr: @malloryfreeberg Thanks for this note. To my mind "end bias" implies that more of the aligned reads are near the 3' end or 5' end. "3 prime tag" and "5 prime tag" in contrast, imply that the library preparation uses a chemistry that _requires_ the read to be aligned to either the poly-A tail or 5' cap. 

One could envision a continuum of localization from the start to end of the transcript that is: `5 prime tag > 5 prime bias > no bias > 3 prime bias > 3 prime tag`. 

The only valid value for 10x v2 should be `3 prime tag`. For the 5'+TCR 10x sequencing kit the only valid value should be `5 prime tag`. 

Bulk RNA-seq captured with a poly-A pull down and rolling circle amplified (has an appreciable polymerase fall-off rate and reverse transcription is initiated at the 3' end) could be `3 prime biased`. I don't have an example of an assay that would be 5' biased off the top of my head.

scripts/vx_queries/10x-query.json
ambrosejcarr: Looking forwards, I am aware of large numbers of people using the commercially available reagents alongside the 10x chromium chip. I imagine it will be necessary at some point in the future to understand how to categorize the assays on this page: 

https://www.10xgenomics.com/solutions/

cF. 10x CITEseq: https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/feature-bc

scripts/vx_queries/10x-query.json
malloryfreeberg: Ah, this is great to know! Ok, so 10x 3' v2 should _always_  be `3 prime tag`. Got it. Then yes, this ES query match for `end_bias` is good.

samanehsan: Sounds good! Thanks for the clarification!

scripts/vx_queries/10x-query.json
samanehsan: Is there a standard set of `sequencing_approach` values for single cell 10xv2 sequencing? I was looking too see if there was a single value I could match on but found both `EFO:0008995 (10x sequencing)` and `EFO:0008440 (Tag based single cell RNA sequencing)` being used

samanehsan: I'm not sure...if we match on `sequencing_approach.text: "CITE-seq"` that would be more human readable, but I don't think it is guaranteed to be the same across projects so there could be cite-seq or Cite-seq as well.

samanehsan:
@malloryfreeberg @daniwelter could we make the `nucleic_acid_source` field required for `library_preparation_protocol_json`? If a user leaves it out then it won't match our query after this change.

scripts/vx_queries/10x-query.json
samanehsan: @Bento007 @mikebaumann do you know of a way we could do this with elasticsearch? We've been matching on ontology ids since they are reliably consistent, but it's hard to tell what term they map to from looking at the query.

samanehsan: Oh! I just realized...I think we can match on the `sequencing_approach.ontology_label` field here instead which is both human readable and consistent :octocat:

malloryfreeberg:
> @malloryfreeberg @daniwelter could we make the `nucleic_acid_source` field required for `library_preparation_protocol_json`? If a user leaves it out then it won't match our query after this change.

This shouldn't be a problem. I'll make a ticket in the schema repo so we can discuss broadly.

Edit: https://github.com/HumanCellAtlas/metadata-schema/issues/824

rexwangcc (approved):
LGTM, thanks! Related to this PR, we might want to start thinking about adding readmes to the queries, since the JSON files don't have the concept of "comments"

samanehsan:
I noticed an error in the `nucleic_acid_source` field that I had added to the 10x test spreadsheet in integration, so I will wait to create a new subscription until the fix is merged: https://github.com/HumanCellAtlas/metadata-schema/pull/842 :octocat:

PR #134, opened 2019-02-15
Update connexion and standardize api

rexwangcc:
### Purpose
_Please explain the purpose of this PR and include links to any GitHub issues that it fixes:_

- https://app.zenhub.com/workspaces/dcp-backlogs-5ac7bcf9465cb172b77760d9/issues/humancellatlas/secondary-analysis/409
---
### Changes
_Please list out what major changes were made in this PR to address the issue:_

- Update our API ymal to accurately reflect the requests and responses-  models.
- Switch to use a newer version of Connexion App.
- Use the latest Swagger UI bundle.
- We have a swagger UI like this now:
screen shot 2019-02-21 at 12 13 48 pm

---
### Review Instructions
_Please provide instructions about how should a reviewer test/verify the changes in this PR:_

- No instructions.

---
### PR Checklist

---
### Follow-up Discussions
_Please append follow-up discussions and issues during the review process below:_

- No follow-up discussions.


lira/lira_config.py
```diff
@@ -229,7 +229,8 @@ def required_fields(self):
             'version',
             'dss_url',
             'ingest_url',
-            'schema_url'
+            'schema_url',
+            'lira_domain_name'
```
samanehsan: Is this the same as the `DOMAIN` field in the lira config?  `https://github.com/HumanCellAtlas/lira/blob/master/deploy/config_files/config.sh.ctmpl#L25

rexwangcc: Good catch! I didn't realize that! Let me re-use that name in this class...

samanehsan: I think you'll have to add the field to the test config as well so that the unit tests pass: https://github.com/HumanCellAtlas/lira/blob/master/lira/test/data/config.json

rexwangcc:
The unit test just passed, not sure why the Github UI doesn't show it: https://travis-ci.org/HumanCellAtlas/lira/builds/494022486

lira/lira_api.yml
```diff
@@ -0,0 +1,242 @@
...
+        items:
+          type: string
+          format: uuid
+      project_shortname:
+        type: string
+      project_uuid:
+        type: string
+        format: uuid
+  Error:
+    type: object
```
samanehsan, line 150: Since this error is now an object instead of a string, do we need to update all of the places where these api methods return an error message to instead return an object?

lira/lira_config.py
samanehsan: Oh! Also I just realized that variable does not get passed into the lira config template right now (it's only set for updating certs) so that needs to be added as well: https://github.com/HumanCellAtlas/lira/blob/master/deploy/config_files/lira-config.json.ctmpl

lira/lira_api.yml
rexwangcc: Yes, good catch! I've updated the code based to reflect this change.

rhiananthony (approved):
👍

pshapiro4broad (approved):
Looks good, as far as I can tell.

lira/lira_api.yml
```diff
@@ -0,0 +1,242 @@
+swagger: '2.0'
```
pshapiro4broad, line 1: Does this need to be updated to 3.0?

rexwangcc: This is a good question, 3.0 has a lot of great features (e.g. OneOf, AnyOf) for building API definitions, my concerns were that all of the other components are still using v2.0, according to their swagger.json:
- https://dss.data.humancellatlas.org/
- https://upload.data.humancellatlas.org/

So I kept it 2.0 in this PR. Actually I think we should bring this topic up to the DCP Tech Arch meeting!

rexwangcc:
The Mintegration test just passed! https://gotc-jenkins.dsp-techops.broadinstitute.org/job/integration-tests/1739/console

lira/api/notifications.py
```diff
@@ -89,7 +89,13 @@ def post(body):
         )
         if cromwell_response.status_code > 201:
             logger.error("HTTP error content: {content}".format(content=cromwell_response.text))
-            response_json = dict(result=cromwell_response.text)
+            # TODO: Improve the error handling here to make the error msgs more granular
+            response_json = {
+                'detail': cromwell_response.text,
+                'status': 500,
```
samanehsan: Minor suggestion: You can replace the hard-coded status here with the `status_code` variable (after moving it above the `response_json`)

README.md
```diff
@@ -1,8 +1,11 @@
...
+![Travis (.org) branch](https://img.shields.io/travis/HumanCellAtlas/lira/master.svg?label=Unit%20Test&style=flat)
+[![Docker Repository on Quay](https://quay.io/repository/humancellatlas/secondary-analysis-lira/status "Docker Repository on Quay")](https://quay.io/repository/humancellatlas/secondary-analysis-lira)
+![GitHub release](https://img.shields.io/github/release/HumanCellAtlas/lira.svg?label=Latest%20Release&style=flat&colorB=green)
+![Github](https://img.shields.io/badge/python-2.7%20%7C%203.6-green.svg?style=flat&logo=python&colorB=blue)
+![GitHub](https://img.shields.io/github/license/neuralnotworklab/stylegan.svg?style=flat)
+![Github](https://img.shields.io/badge/Slack%20Channel-%23hca--dcp--analysis--community-green.svg?style=flat&?link=https://humancellatlas.slack.com/messages/secondary-analysis/&link=https://humancellatlas.slack.com/messages/secondary-analysis/&colorB=blue)
```
samanehsan: I know this is not quite related to this PR, but could you change the slack channel link to the data-pipelines-team since the channel was renamed?

lira/lira.py
```diff
@@ -20,21 +29,37 @@
...
+
+# Consume the config and initialize the LRU cache for WDL submissions
 app.app.config = config
 app.app.prepare_submission = lira_utils.create_prepare_submission_function(app.app.config.cache_wdls)
 
+
+# Use automatic routing with custom resolver: https://github.com/zalando/connexion#automatic-routing
 resolver = RestyResolver('lira.api', collection_endpoint_name='list')
-app.add_api('lira.yml', resolver=resolver, validate_responses=True)
+arguments = {'API_DOMAIN_NAME': config['DOMAIN']}
```
samanehsan: It looks like this needs to be `config.domain` here instead of `config['DOMAIN']` because `config` is an instance of the `LiraConfig` class.

README.md
rexwangcc: This is a good point, changed it to `https://humancellatlas.slack.com/messages/analysis-community` since it was the channel originally for community talks (was `secondary-analysis` and now `analysis-community`)

lira/lira.py
rexwangcc: I think both would work but you are right, `config.domain` is more formal, changed.

lira/lira_api.yml
rexwangcc: As discussed in the DCP Tech Arch meeting, nothing is blocking us from upgrading to OpenAPI3, but I'd prefer to do that in a separate PR

samanehsan (approved):

rexwangcc:
retest this please jenkins

PR #161, opened 2019-05-01
Subscription query migration

rexwangcc:
### Purpose
_Please explain the purpose of this PR and include links to any GitHub issues that it fixes:_

- https://broadinstitute.atlassian.net/jira/software/projects/GH/boards/530?selectedIssue=GH-204
---
### Changes
_Please list out what major changes were made in this PR to address the issue:_

- Migrate all ES queries to JMESPath.
- Update the subscribe.py.
---
### Review Instructions
_Please provide instructions about how should a reviewer test/verify the changes in this PR:_

- No instructions.

---
### PR Checklist

---
### Follow-up Discussions
_Please append follow-up discussions and issues during the review process below:_

- No follow-up discussions.


scripts/jmespath_queries/smartseq2.jmespath
```diff
@@ -0,0 +1,5 @@
+(files.library_preparation_protocol_json[].library_construction_method[].ontology | contains(@, `EFO:0008931`))
+ && (files.sequencing_protocol_json[].paired_end | [0])
+ && files.donor_organism_json[].biomaterial_core.ncbi_taxon_id[] | (min(@) == `9606` && max(@) == `9606`)
+ && files.sequencing_protocol_json[].sequencing_approach.ontology_label | not_null(@, `[]`) | !contains(@, `CITE-seq`)
+ && files.analysis_process_json[].process_type.text | not_null(@, `[]`) | !contains(@, `analysis`)
```
samanehsan: `process_type` is now `type`

scripts/jmespath_queries/10x.jmespath
```diff
@@ -0,0 +1,6 @@
+(files.library_preparation_protocol_json[].library_construction_method[].ontology_label | contains(@, `10X v2 sequencing`))
+ && (files.library_preparation_protocol_json[].end_bias | contains(@, `3 prime tag`))
+ && (files.library_preparation_protocol_json[].nucleic_acid_source | contains(@, `single cell`))
+ && files.donor_organism_json[].biomaterial_core.ncbi_taxon_id[] | (min(@) == `9606` && max(@) == `9606`)
+ && files.sequencing_protocol_json[].sequencing_approach.ontology_label | not_null(@, `[]`) | !contains(@, `CITE-seq`)
+ && files.donor_organism_json[].biomaterial_core.ncbi_taxon_id | not_null(@, `[]`) | !contains(@, `analysis`)
```
samanehsan: I think this is missing "&& files.analysis_process_json[].process_type.text | not_null(@, `[]`)" before the "| !contains(@, `analysis`"

scripts/jmespath_queries/metadata_attachments.json
```diff
@@ -0,0 +1,14 @@
+{
+  "project_shortname": {
```
samanehsan, line 2: This file looks out of date because it's missing `project_uuid` and it has `submitter_id`. You should be able to use this one from the master branch: https://github.com/HumanCellAtlas/lira/blob/master/scripts/vx_queries/metadata_attachments.json

scripts/jmespath_queries/smartseq2.jmespath
samanehsan: Was the "should" clause removed because it wasn't really impacting the query results? https://github.com/HumanCellAtlas/lira/blob/master/scripts/vx_queries/smartseq2-query.json#L22

rexwangcc: good catch!

scripts/jmespath_queries/10x.jmespath
rexwangcc: Good catch, and i guess here the `process_type ` would be type?

scripts/subscribe.py
```diff
@@ -35,14 +40,23 @@ def create_subscription(args):
...
-    payload = _prep_json(
-        callback_base_url=args['callback_base_url'],
-        query_json_file=args['query_json'],
-        query_param_token=args.get('query_param_token'),
-        hmac_key_id=args.get('hmac_key_id'),
-        hmac_key=args.get('hmac_key'),
-        attachments_file=args['additional_metadata'],
-    )
+    if args['subscription_type'] == 'elasticsearch':
+        warnings.warn(
```
samanehsan: Does this throw an exception?

scripts/jmespath_queries/10x.jmespath
rexwangcc: Actually by re-visiting it looks like the whole last line doesn't match the es query, @xbrianh could you give any insights here?

scripts/jmespath_queries/metadata_attachments.json
rexwangcc: Done, thanks for catching it!

scripts/jmespath_queries/10x.jmespath
samanehsan: Oh I see what you mean...the `biomaterial_core.ncbi_taxon_id` filter is also on line 4. I think the last line needs to be replaced with:
```
&& files.analysis_process_json[].type.text | not_null(@, `[]`) | !contains(@, `analysis`)
```

scripts/subscribe.py
rexwangcc: No, this will throw a warning (a stdout message that first time this is called)

samanehsan: Oh okay! So if I specify `subscription_type` == elasticsearch, the payload does not get defined: https://github.com/HumanCellAtlas/lira/pull/161/files/c52bd0996ea997dac8c0e2c5d3eee7b40acbc4ec#diff-a84b26f5ad176c0a31ef9df2ed4e43cbR66

So think it makes sense to either throw an exception here so that the script fails before it gets to that point, or throw a warning but still allow creating an elasticsearch subscription.

scripts/jmespath_queries/10x.jmespath
xbrianh: @rexwangcc @samanehsan Good catch, sorry for the typo :/

scripts/subscribe.py
```diff
@@ -4,12 +4,17 @@
...
 try:
     from DCPAuthClient import DCPAuthClient
 except ModuleNotFoundError:
     print('Cannot find the DCPAuthClient within this folder!')
-    exit(1)
+    sys.exit(1)
+
+if sys.version_info[0] < 3:
+    print('You need to run this with Python 3')
+    sys.exit(1)
```
tbl3rd: Thank you for this, but can you just?

sys.exit('You need to run this with Python 3.')

Similarly elsewhere.

scripts/subscribe.py
```diff
@@ -64,35 +78,44 @@ def create_subscription(args):
...
         with open(attachments_file) as f:
             attachments = json.load(f)
         payload['attachments'] = attachments
 
+    if not jmespath_query:
+        with open(query_file) as f:
+            query = json.load(f)
+        payload['es_query'] = query
+    else:
+        query = ""
```
tbl3rd: Lift the `query` initialization out of the conditional?

scripts/subscribe.py
```diff
@@ -64,35 +78,44 @@ def create_subscription(args):
...
+        with open(query_file, 'r') as f:
```
tbl3rd: Why `'r'` here and not above?

scripts/subscribe.py
```diff
@@ -64,35 +78,44 @@ def create_subscription(args):
...
-            'callback_url': '{0}?auth={1}'.format(callback_base_url, query_param_token),
+            'callback_url': '{0}?auth={1}'.format(callback_base_url, query_param_token)
         }
 
     if attachments_file:
         with open(attachments_file) as f:
             attachments = json.load(f)
         payload['attachments'] = attachments
 
+    if not jmespath_query:
```
tbl3rd: Maybe invert the sense of this conditional and dispense with the not?

if jmespath_query:

scripts/subscribe.py
```diff
@@ -64,35 +78,44 @@ def create_subscription(args):
...
+        payload['es_query'] = query
+    else:
+        query = ""
+        with open(query_file, 'r') as f:
+            for line in f:
+                query += line.strip('\n')
+
+        # TODO: find a way to validate the JMESPath query
+        payload['jmespath_query'] = query
     return payload
```
tbl3rd, line 163: Maybe just ...
```
    with open(query_file) as f:
        if jmespath_query:
            query = ""
            for line in f:
                query += line.strip('\n')
            payload['jmespath_query'] = query
        else:
            payload['es_query'] = json.load(f)
```
... but I also don't understand stripping newlines in one branch.
In other words, why not just:
```
    query = ""
    with open(query_file) as f:
        for line in f:
            query += line.strip()
    payload['jmespath_query' if jmespath_query else 'es_query'] = query
```

scripts/subscribe.py
```diff
@@ -175,7 +198,7 @@ def parser(arguments):
         sub_command.add_argument(
             '--subscription_type',
             help='Which type of subscription query you want to use, "elasticsearch" by default. ["elasticsearch", "jmespath"]',
-            default='elasticsearch',
+            default='jmespath',
```
tbl3rd, line 249: Maybe split the `_prep_json` proc in two: one for jmespath and one for Es?
Dispatch to them from `create_subscription()` based on `args` there?
Then deprecate the Es one.

scripts/subscribe.py
```diff
@@ -35,14 +40,23 @@ def create_subscription(args):
...
+        )
+    else:
+        payload = _prep_json(
+            callback_base_url=args['callback_base_url'],
+            query_file=args['query_file'],
+            query_param_token=args.get('query_param_token'),
+            hmac_key_id=args.get('hmac_key_id'),
+            hmac_key=args.get('hmac_key'),
+            attachments_file=args['additional_metadata'],
+            jmespath_query=True,
```
tbl3rd: Adding a Boolean argument like this always feels to me like trying to mix 2 functions into one.
Why not keep the 2 code paths separate?

tbl3rd (commented):
Just commenting. Cannot validate the queries without help.

scripts/subscribe.py
rexwangcc: Good point! I think I meant to go the latter way but somehow failed :(

scripts/subscribe.py
rexwangcc: add it to all of the places that I've missed, `r` == `read`, which should be explicitly defined.

scripts/subscribe.py
rexwangcc: > In other words, why not just:

Since the JMESPATH query is plain text which has `\n` at the end while ELasticSearch queries are valid JSON.

scripts/README.md
```diff
@@ -1,13 +1,63 @@
+## Subscription queries
+
+The communication between the [HCA DCP Data Store Service](https://github.com/HumanCellAtlas/data-store) and Lira is 
+relying on the subsciptions. This folder contains the necessary scripts to create/list/delete subscriptions in Data Store
+Service for Lira:
+
+- `DCPAuthClient.py`: handles the authentication with HCA DCP Data Store, requires Python 3.x to run.
+- `subscribe.py`: is the core client script for working with subscriptions, requires Python 3.x to run.
+- `v*_queries` and `jmespath_queries`: are the directories that host subscription queries.
```
samanehsan, line 9: I think we can delete the v6 queries at some point since we no longer use them (doesn't have to be in this PR)

scripts/README.md
```diff
@@ -1,13 +1,63 @@
...
+To create a new subscription, run:
+
+```shell
+bash subscription-create.sh config-dev.sh jmespath_queries/smartseq2.jmespath jmespath_queries/metadata_attachments.json
+``` 
+
+here we assume:
+- `config-dev.sh` is the config file you created following the `config-template.sh` for your `dev` environment to subscribe to `dev` Data Store.
+- `jmespath_queries/smartseq2.jmespath` is the query for Data Store to determine whether it should notify Lira when a new bundle is indexed on their side.
+- `jmespath_queries/metadata_attachments.json` is a call-back based attachment file specifying all of the necessary meta fields for a bundle that Lira requires from Data Store to submit a workflow.
```
samanehsan: nit: `meta` --> `metadata`

scripts/README.md
```diff
@@ -1,13 +1,63 @@
...
+```shell
+bash subscriptions-get.sh config-dev.sh
+```
+
+#### Delete a subscription
+
+To delete a subscription, run:
+
+```shell
+bash subscription-delete.sh config-dev.sh bfc5eeba-8a3d-4ce7-8905-4ae314f8e210
```
samanehsan: Is this a real subscription id? Maybe replace it with `<subscription_id>` instead

scripts/subscribe.py
```diff
@@ -58,41 +77,75 @@ def create_subscription(args):
...
-    with open(query_json_file) as f:
+    if hmac_key_id is not None:
+        print('Subscribing with HMAC key')
+        payload = {
+            'callback_url': callback_base_url,
+            'hmac_key_id': hmac_key_id,
+            'hmac_secret_key': hmac_key,
+        }
+    else:
+        print('Subscribing with query param token')
```
samanehsan, line 103: Does subscribing with a query param token work anymore? If not we can simplify this script by removing it.

scripts/subscribe.py
```diff
@@ -58,41 +77,75 @@ def create_subscription(args):
...
     callback_base_url,
-    query_json_file,
+    query_file,
     query_param_token,
     hmac_key_id,
     hmac_key,
     attachments_file,
 ):
-    with open(query_json_file) as f:
+    if hmac_key_id is not None:
```
samanehsan, line 92: I know this is existing code, but checking `if hmac_key_id and hmac_key` would be useful here since they are both required in order for hmac auth to work

scripts/subscribe.py
samanehsan: If we do keep this in, I suggest changing `else` --> `elif query_param_token` (otherwise you can accidentally set a callback url with `None` as the token) and then adding an `else` clause at the end that raises an error

scripts/jmespath_queries/smartseq2.jmespath
```diff
@@ -0,0 +1,5 @@
+(files.library_preparation_protocol_json[].library_construction_method[].ontology | contains(@, `EFO:0008931`))
+ && (files.sequencing_protocol_json[].paired_end | [0])
+ && files.donor_organism_json[].biomaterial_core.ncbi_taxon_id[] | (min(@) == `9606` && max(@) == `9606`)
+ && files.sequencing_protocol_json[].sequencing_approach.ontology_label | not_null(@, `[]`) | !contains(@, `CITE-seq`)
+ && files.analysis_process_json[].type.text | not_null(@, `[]`) | !contains(@, `analysis`)
```
samanehsan: nit: newline

scripts/jmespath_queries/smartseq2.jmespath
```diff
@@ -0,0 +1,5 @@
+(files.library_preparation_protocol_json[].library_construction_method[].ontology | contains(@, `EFO:0008931`))
+ && (files.sequencing_protocol_json[].paired_end | [0])
```
samanehsan: Is this checking that `paired_end` is true?

samanehsan:
Left a few questions/comments otherwise this looks good to me :octocat:

scripts/README.md
rexwangcc: removed!

scripts/README.md
rexwangcc: Using a Nil UUID as @tbl3rd suggested `00000000-0000-0000-0000-000000000000`

scripts/subscribe.py
rexwangcc: Good point! Will do

scripts/jmespath_queries/smartseq2.jmespath
rexwangcc: I believe so, it is using pipe `|` to get the first (and the only) value of the `paired_end `, if there is no such value, the result would be `Null` and the query will fail.

rexwangcc: ^ My guess based on http://jmespath.org/tutorial.html#pipe-expressions, @xbrianh does it make sense?

rexwangcc: Actually, I misunderstood @samanehsan 's question. The query seems to be checking the existence of the `paired_end` key, not the value of the field. @xbrianh is it true?

samanehsan:
I think this is good to merge once we resolve https://github.com/HumanCellAtlas/lira/pull/161#discussion_r282640118 :octocat:

scripts/jmespath_queries/smartseq2.jmespath
rexwangcc: Bump @xbrianh

xbrianh: @rexwangcc Yes, that's true, but there's a wrinkle: the value for `paired_end` is a boolean.

rexwangcc:
@xbrianh I'm merging this PR for now. Since the pipeline-surge jamboree added a lot more pipelines(which require more subscription queries), I propose we hold off the migration for a week, until we finished writing the queries for them.

scripts/jmespath_queries/smartseq2.jmespath
rexwangcc: @xbrianh Let me double check if I understand your comment! Based on my experiments on http://jmespath.org/tutorial.html:

`sequencing_protocol_json[].paired_end | [0]` will:
- return `false` on: `{"sequencing_protocol_json": [{"paired_end": false}]}`
- return `null` on: `{"sequencing_protocol_json": [{"single_end": true}]}`
- return `true` on: `{"sequencing_protocol_json": [{"paired_end": true}]}`
- return `false` on: `{"sequencing_protocol_json": [{"paired_end": false}]}`

which means, since the value for `paired_end` is a boolean, it doesn't really matter to check the existence or the value, cauz they acts the same way?

xbrianh: Yes, that is my finding. However, perhaps it's a good idea to be explicit:
```
In [18]: jmespath.search("sequencing_protocol_json[].paired_end | [0]==`true`", {"sequencing_protocol_json": [{"paired_end": True}]})                                             
Out[18]: True

In [19]: jmespath.search("sequencing_protocol_json[].paired_end | [0]==`false`", {"sequencing_protocol_json": [{"paired_end": True}]})                                            
Out[19]: False

In [20]: jmespath.search("sequencing_protocol_json[].paired_end | [0]==`25`", {"sequencing_protocol_json": [{"paired_end": True}]})                                               
Out[20]: False
```

rexwangcc: Cool, changed!

samanehsan (approved):

tbl3rd (approved):

PR #163, opened 2019-05-14
Remove `should` clauses from lira queries

mckinsel:
They don't do anything, they just look like they do. "If the bool query
is in a query context and has a must or filter clause then a document
will match the bool query even if none of the should queries match."

https://www.elastic.co/guide/en/elasticsearch/reference/6.3/query-dsl-bool-query.html

### DIFF
--- a/subscription/elasticsearch_queries/smartseq2-query.json
+++ b/subscription/elasticsearch_queries/smartseq2-query.json
@@ -19,18 +19,6 @@
           }
         }
       ],
-      "should": [
-        {
-          "match": {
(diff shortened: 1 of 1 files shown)


samanehsan (approved):

samanehsan:
Thanks for the PR Marcus! I think what we should add to the query is to match on `files.dissociation_protocol_json.method.ontology` for plate-based smartseq2, or exclude fluidigm-based smartseq2. @daniwelter do we have either of these ontology terms for dissociation protocol?

samanehsan:
retest this please

rhiananthony (approved):

PR #167, opened 2019-06-10
Label every workflow with a hash of its inputs

samanehsan:
### Purpose
Label every workflow with a hash of its inputs so that it is possible to check if a workflow has already been run on that data.

### Changes
Label SS2 and Optimus workflows with a hash of the sample_id, taxon_id, and fastq file hashes. For each fastq file, concatenate the sha1, sha256, s3-etag and crc32c hashes provided by the data store. 

Note: Because the default digest size is 32, and the hexdigest is twice as long, each label will always be 64 characters long (which is below Cromwell's cut-off limit of 255 characters) 

### Review Instructions
Confirm that two workflows run on the exact same bundle produce the same label hash by checking the "labels" tab in job manager and looking for "hash-id".

Example for SmartSeq2:
1) https://job-manager.caas-prod.broadinstitute.org/jobs/e466e536-9cf3-4089-a870-76c0b4aca72e
2) https://job-manager.caas-prod.broadinstitute.org/jobs/4e319415-e9ae-4216-993d-2f9395204064

### DIFF
--- a/lira/bundle_inputs.py
+++ b/lira/bundle_inputs.py
@@ -2,37 +2,64 @@ import hashlib
 from pipeline_tools.shared import metadata_utils, tenx_utils, http_requests
 
 
-def get_smartseq2_workflow_inputs(bundle):
+def get_hashes_from_file_manifest(file_manifest):
+    """ Return a string that is a concatenation of the file hashes provided in the bundle manifest entry for a file:
+        {sha1}{sha256}{s3_etag}{crc32c}
+    """
+    sha1 = file_manifest.sha1
+    sha256 = file_manifest.sha256
+    s3_etag = file_manifest.s3_etag
+    crc32c = file_manifest.crc32c
+    return '{sha1}{sha256}{s3_etag}{crc32c}'.format(sha1=sha1, sha256=sha256, s3_etag=s3_etag, crc32c=crc32c)
+
+
+def get_smartseq2_workflow_input_hash(bundle):
     """ Return a string that is a concatenation of bundle-specific Smartseq2 paired-end workflow inputs:
-        {sample_id}{taxon_id}{read_1_sha256}{read2_sha256}
+        {sample_id}{taxon_id}{read_1}{read2}
+
+        Where each "file hash" is a concatenation of {sha1}{sha256}{s3_etag}{crc32c}
     """
     sample_id = metadata_utils.get_sample_id(bundle)
     taxon_id = metadata_utils.get_sample_id(bundle)
-    read_1_sha = [sf.manifest_entry.sha256 for sf in bundle.sequencing_output if sf.read_index == 'read1']
-    read_2_sha = [sf.manifest_entry.sha256 for sf in bundle.sequencing_output if sf.read_index == 'read2']
+    r1_manifest = [sf.manifest_entry for sf in bundle.sequencing_output if sf.read_index == 'read1']
+    r2_manifest = [sf.manifest_entry for sf in bundle.sequencing_output if sf.read_index == 'read2']
+    r1_hashes = get_hashes_from_file_manifest(r1_manifest[0])
+    r2_hashes = get_hashes_from_file_manifest(r2_manifest[0])
     return bytes('{sample_id}{taxon_id}{read_1}{read_2}'.format(sample_id=sample_id,
                                                                 taxon_id=taxon_id,
-                                                                read_1=read_1_sha[0],
-                                                                read_2=read_2_sha[0]), encoding='utf8')
+                                                                read_1=r1_hashes,
+                                                                read_2=r2_hashes), encoding='utf8')
 
 
-def get_optimus_workflow_inputs(bundle):
+def get_optimus_workflow_input_hash(bundle):
+    """ Return a string that is a concatenation of bundle-specific Optimus workflow inputs
+    where the file hashes are for each set of read_1, read_2 and index_1 files in the order
+    of their flow cell lane number. For example:
+    {sample_id}{taxon_id}{lane1_r1}{lane1_r2}{lane1_i1}{lane2_r1}{lane2_r2}{lane2_i1}
+
+    Each "file hash" is a concatenation of {sha1}{sha256}{s3_etag}{crc32c}
+    """
     sample_id = metadata_utils.get_sample_id(bundle)
     taxon_id = metadata_utils.get_sample_id(bundle)
     inputs = '{sample_id}{taxon_id}'.format(sample_id=sample_id, taxon_id=taxon_id)
     fastq_files = [f for f in bundle.files.values() if str(f.format).lower() in ('fastq.gz', 'fastq')]
     fastq_dict = tenx_utils.create_fastq_dict(fastq_files)
-    for lane in fastq_dict:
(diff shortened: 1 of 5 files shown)


lira/bundle_inputs.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    sha1 = file_manifest.sha1
+    sha256 = file_manifest.sha256
+    s3_etag = file_manifest.s3_etag
+    crc32c = file_manifest.crc32c
+    return '{sha1}{sha256}{s3_etag}{crc32c}'.format(sha1=sha1, sha256=sha256, s3_etag=s3_etag, crc32c=crc32c)
+
+
+def get_smartseq2_workflow_input_hash(bundle):
+    """ Return a string that is a concatenation of bundle-specific Smartseq2 paired-end workflow inputs:
```
rexwangcc: nit: this says `string` while the function is actually returning a `bytes`

lira/bundle_inputs.py
```diff
@@ -0,0 +1,71 @@
...
+
+        Where each "file hash" is a concatenation of {sha1}{sha256}{s3_etag}{crc32c}
+    """
+    sample_id = metadata_utils.get_sample_id(bundle)
+    taxon_id = metadata_utils.get_sample_id(bundle)
+    r1_manifest = [sf.manifest_entry for sf in bundle.sequencing_output if sf.read_index == 'read1']
+    r2_manifest = [sf.manifest_entry for sf in bundle.sequencing_output if sf.read_index == 'read2']
+    r1_hashes = get_hashes_from_file_manifest(r1_manifest[0])
+    r2_hashes = get_hashes_from_file_manifest(r2_manifest[0])
+    return bytes('{sample_id}{taxon_id}{read_1}{read_2}'.format(sample_id=sample_id,
```
rexwangcc: optional: since lira is now fully migrated to Py3.6, using f-strings here could make the string formatting a bit cleaner. (e.g. `bytes(f'{sample_id}{taxon_id}{read_1_hashes}{read_2_hashes}', encoding='utf8')`), same to other places below

lira/bundle_inputs.py
rexwangcc: Also, if this function is `paired-end` specific, could we modify the function name to be accurate?

lira/bundle_inputs.py
```diff
@@ -0,0 +1,71 @@
...
+        r2_hashes = get_hashes_from_file_manifest(fastq_dict[lane]['read2'])
+        file_hashes = '{0}{1}'.format(r1_hashes, r2_hashes)
+        if fastq_dict[lane].get('index1'):
+            i1_hashes = get_hashes_from_file_manifest(fastq_dict[lane]['index1'])
+            file_hashes += i1_hashes
+        inputs += file_hashes
+    return bytes(inputs, encoding='utf8')
+
+
+WORKFLOW_INPUTS = {
```
rexwangcc, line 5: It might be better to put this mapping at the top of the file

lira/bundle_inputs.py
```diff
@@ -0,0 +1,71 @@
...
+    'AdapterSmartSeq2SingleCell': get_smartseq2_workflow_input_hash,
+    'AdapterOptimus': get_optimus_workflow_input_hash
+}
+
+
+def create_workflow_inputs_hash_label(workflow_name, bundle_id, bundle_version, dss_url):
+    get_inputs_fn = WORKFLOW_INPUTS.get(workflow_name, None)
+    if get_inputs_fn:
```
rexwangcc: This means the workflows that are not in the "supported mapping" `WORKFLOW_INPUTS` will be ignored and won't get a hash. I think we need to discuss how we wanna handle that case. Do we really want to start the workflows that doesn't have a deterministic set of inputs via Lira? If so, what are the use cases? If not, can we just raise an error here?

rexwangcc (commented):
It looks great! Besides the comments below I have 3 more general questions:
1. I saw we used `workflow-hash` as the key of this label. This is somewhat confusing to me since it implies "hash of the workflow" to some extent. Could we use a more specific word for it? Something like, `workflow_input_digest` or `workflow_input_hash`...
2. There is some code in `bundle_inputs.py` that extract the metadata files from a bundle instance and then calculate the hashes. This piece of code might grow as we have more and more pipelines supported by Lira in the future. Ideally, it would be great if Lira doesn't have to know the `mandatory inputs` of each pipeline, instead, it calls pipeline-tools to get a manifest of the inputs given the pipeline name, and then calculates the hashed, so we won't have 2 places(lira and pipeline-tools) storing the pipeline inputs information. What do you think?
3. Can we have some tests for the `bundle_inputs.py`?

lira/bundle_inputs.py
```diff
@@ -1,71 +1,21 @@
...
+    'AdapterSmartSeq2SingleCell': smartseq2.get_ss2_paired_end_inputs_to_hash,
+    'AdapterOptimus': optimus.get_optimus_inputs_to_hash
 }
 
 
 def create_workflow_inputs_hash_label(workflow_name, bundle_id, bundle_version, dss_url):
+    """
+    Create a hash out of the bundle-specific inputs for a workflow.
+    """
     get_inputs_fn = WORKFLOW_INPUTS.get(workflow_name, None)
```
rexwangcc: It'd be better to find a good name for `get_inputs_fn ` since it's hash than function name now.

lira/bundle_inputs.py
```diff
@@ -1,71 +1,21 @@
...
     if get_inputs_fn:
-        bundle = metadata_utils.get_bundle_metadata(bundle_id, bundle_version, dss_url, http_requests.HttpRequests())
-        workflow_inputs = get_inputs_fn(bundle)
-        sha256_hash = hashlib.sha256(workflow_inputs)
+        workflow_inputs = get_inputs_fn(bundle_id, bundle_version, dss_url)
+        formatted_inputs = ''
+        for i in workflow_inputs:
+            formatted_inputs += i
```
rexwangcc: would it be possible to use something like `''.join(workflow_inputs)` to save lines 🤔 ?

rexwangcc (commented):
Looks good!

lira/bundle_inputs.py
samanehsan: How about `get_inputs_to_hash`?

rexwangcc: Sounds good!

rexwangcc (approved):

requirements.txt
```diff
@@ -1,16 +1,15 @@
...
-requests-mock==1.5.0
+requests-mock>=1.5.2, <2
 oauth2client==4.1.2
 mock==2.0.0
 requests-http-signature==v0.0.3
 cromwell-tools>=1.1.1, <2
 black==19.3b0
 flake8==3.7.7
 pre-commit==1.14.4
+git+git://github.com/HumanCellAtlas/pipeline-tools@70c859b2c90ec95c98dcd15d679a1f2fe99b30be
```
samanehsan: I will update this with a released version of pipeline-tools before merging! ☀️

samanehsan:
I manually ran the integration tests and linked them in the PR description, since Quay is throwing errors when trying to build the Lira image for the branch which has "/" in it's name.

PR #172, opened 2019-07-25
Fix linting tests

samanehsan:
### Purpose
The `scripts` directory was deleted, so the formatting check `black lira/ scripts/ --skip-string-normalization --check` throws an error

### Changes
Remove the `scripts` directory from the linting check

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,9 @@ jobs:
     - stage: Linting Test
       script: 
         # Check Black code style compliance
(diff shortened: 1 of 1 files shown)


rexwangcc (approved):
Oops, good catch!

rexwangcc:
Actually, we could even remove the `lira` so it also checks the `subscription` folder for us!

samanehsan:
Good point! I forgot that folder had some python scripts in it as well

PR #174, opened 2019-07-25
Make subscriptions to test data more specific

samanehsan:
The subscription queries need to be pipeline-specific so that the correct pipeline is run off of the
notification!


rexwangcc (approved):
Looks good!

PR #175, opened 2019-07-30
Revert "Label every workflow with a hash of its inputs (#167)"

samanehsan:
### Purpose
The change to add a hash label to workflows in Lira causes a timeout issue when responding to DSS notifications. We reverted the version of Lira used in our deployments, but we should revert this commit as well so that we can continue releasing Lira code. 

### Changes
This reverts commit 232d1f6cdbee0f21a17c0debd9908b1d02645acf.


.travis.yml
```diff
@@ -3,10 +3,6 @@ language: python
 python:
   - '3.6'
 
-before_install:
-  - pip install --upgrade pip
-  - pip install --upgrade setuptools
-
```
rexwangcc, line 9: Should we remove these?

rexwangcc: Hmm, the tests are passing so I guess we should?

rexwangcc (approved):

.travis.yml
samanehsan: Yeah, I added this to fix issues with installing pipeline-tools so I think it's okay to remove these for now

PR #176, opened 2019-08-13
Modifying subscriptions to DSS to filter bundles such that bundles th…

maniarathi:
…at have a version that is less than or equal to the current latest metadata schema versions are accepted or if the schema version is not populated. Also added a script to generate the query.

### Purpose

The purpose of this PR is to enable the metadata schema unfreeze so that schema changes can be pushed to production without breaking downstream components.

https://app.zenhub.com/workspaces/dcp-5ac7bcf9465cb172b77760d9/issues/humancellatlas/secondary-analysis/631

### Changes

- Secondary Analysis will not be notified when bundles are created including metadata above a version that SA is currently capable of handling including via Metadata API dependencies.

### Review Instructions

- Please test that you are able to receive bundle with ID 1db5775c-a8bd-4f6f-acb6-9233b129f404 in integration environment.

### DIFF
--- a/subscription/elasticsearch_queries/scripts/create_dss_subscription.py
+++ b/subscription/elasticsearch_queries/scripts/create_dss_subscription.py
@@ -74,6 +74,15 @@ def generate_elastic_search_query_for_schema_versions():
       "should": [
         {
           "bool": {
+            "must_not": {
+              "exists": {
+                "field": "files.%s_json.provenance.schema_major_version"
+              }
+            }
+          }
+        },
+        {
+          "bool": {
             "must": [
               {
                 "term": {
@@ -89,6 +98,7 @@ def generate_elastic_search_query_for_schema_versions():
   },
             """
             % (
+                schema_name.value,
                 schema_name.value,
                 LATEST_SUPPORTED_MD_SCHEMA_VERSIONS[schema_name]['major'],
             )
--- a/subscription/elasticsearch_queries/scripts/create_dss_subscription.py
+++ b/subscription/elasticsearch_queries/scripts/create_dss_subscription.py
@@ -1,0 +1,217 @@
+# Script to generate the portion of the DSS subscription via ElasticSearch query based on the last known schema
+# versions that Secondary Analysis is compatible with including compatibility dependencies on Metadata API.
+
+from enum import Enum
+
+
+class MetadataSchemaName(Enum):
+    DONOR_ORGANISM = "donor_organism"
+    SPECIMEN_FROM_ORGANISM = "specimen_from_organism"
+    CELL_SUSPENSION = "cell_suspension"
+    CELL_LINE = "cell_line"
+    ORGANOID = "organoid"
+    IMAGED_SPECIMEN = "imaged_specimen"
+    ANALYSIS_FILE = "analysis_file"
+    REFERENCE_FILE = "reference_file"
+    SEQUENCE_FILE = "sequence_file"
+    SUPPLEMENTARY_FILE = "supplementary_file"
+    IMAGE_FILE = "image_file"
+    PROTOCOL = "protocol"
+    ANALYSIS_PROTOCOL = "analysis_protocol"
+    AGGREGATE_GENERATION_PROTOCOL = "aggregate_generation_protocol"
+    COLLECTION_PROTOCOL = "collection_protocol"
+    DIFFERENTIATION_PROTOCOL = "differentiation_protocol"
+    DISSOCIATION_PROTOCOL = "dissociation_protocol"
+    ENRICHMENT_PROTOCOL = "enrichment_protocol"
+    IPSC_INDUCTION_PROTOCOL = "ipsc_induction_protocol"
+    IMAGING_PROTOCOL = "imaging_protocol"
+    LIBRARY_PREPARATION_PROTOCOL = "library_preparation_protocol"
+    SEQUENCING_PROTOCOL = "sequencing_protocol"
+    IMAGING_PREPARATION_PROTOCOL = "imaging_preparation_protocol"
+    PROJECT = "project"
+    PROCESS = "process"
+    ANALYSIS_PROCESS = "analysis_process"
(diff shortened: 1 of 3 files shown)


subscription/elasticsearch_queries/scripts/create_dss_subscription.py
```diff
@@ -0,0 +1,243 @@
...
+        {
+          "bool": {
+            "must_not": {
+              "exists": {
+                "field": "files.%s_json.provenance.schema_major_version"
+              }
+            }
+          }
+        },
+        {
```
samanehsan, line 84: Can you clarify what this second range clause is doing?

maniarathi: This allows bundles to be accepted if there is no schema version populated (i.e. secondary bundles, older bundles).

subscription/elasticsearch_queries/scripts/create_dss_subscription.py
```diff
@@ -0,0 +1,243 @@
...
+          "bool": {
+            "must": [
+              {
+                "term": {
+                  "files.%s_json.provenance.schema_major_version": %d
+                }
+              },
+              {
+                "range": {
+                  "files.%s_json.provenance.schema_minor_version": {
```
samanehsan: I'm concerned that including the minor schema versions in the query will exclude a lot of bundles and require us to update the query more frequently. Because minor schema updates are not breaking changes, what do you think about taking it out?

maniarathi: Hmm. I am worried more about the dependency against Metadata API. What do you think about leaving it in and then taking it out on a future date if we notice that too many bundles are being excluded?

maniarathi: Also just vaguely looking at the list of changes, there don't seems to be as many minor changes as major changes in general so it might be fine to leave it in for now and remove it once metadata changes settle down into more minor changes? Up to you though!

subscription/elasticsearch_queries/scripts/create_dss_subscription.py
samanehsan: I get including bundles that do not have a schema version populated, but I don't think we can support all of the old major bundle versions 🤔

samanehsan: As discussed over slack: 
> The problem with receiving all old bundles is that we can't actually process the really old versions anymore, hence us wanting to pin to just the versions we can process
> Since we've been in a metadata version freeze, I think we can pin to the latest major version that's in production

subscription/elasticsearch_queries/scripts/create_dss_subscription.py
samanehsan: Hmmm...I'd prefer to take it out to allow us to receive more bundles and then we can add it in if something breaks 🙃 My reasoning here is that any schema update that results in a minor version change shouldn't throw errors in the metadata-api. But maybe @hannes-ucsc can confirm for us?

maniarathi: Fair enough! Taken out :)

subscription/elasticsearch_queries/scripts/create_dss_subscription.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+class MetadataSchemaName(Enum):
+    DONOR_ORGANISM = "donor_organism"
+    SPECIMEN_FROM_ORGANISM = "specimen_from_organism"
+    CELL_SUSPENSION = "cell_suspension"
+    CELL_LINE = "cell_line"
+    ORGANOID = "organoid"
+    IMAGED_SPECIMEN = "imaged_specimen"
+    ANALYSIS_FILE = "analysis_file"
```
samanehsan: We can remove `ANALYSIS_FILE`, `ANALYSIS_PROTOCOL` and `ANALYSIS_PROCESS` because those files will not be present in primary bundles (which is what our query specifies)

samanehsan: I can do this!

samanehsan:
> Please test that you are able to receive bundle with ID 1db5775c-a8bd-4f6f-acb6-9233b129f404 in integration environment.

@maniarathi I'm not getting any results with either the SS2 or 10x query in the integration data store 🙀

maniarathi:
> > Please test that you are able to receive bundle with ID 1db5775c-a8bd-4f6f-acb6-9233b129f404 in integration environment.
> 
> @maniarathi I'm not getting any results with either the SS2 or 10x query in the integration data store 🙀

Checking...

maniarathi:
Ah! Looks like the bundle has been deleted! Creating a new one.

maniarathi:
New bundle in integration available! c66299ee-3ec8-452a-90c7-81a94ac2c0b0

rexwangcc (approved):
It looks good to me. 

My only concern is given we know DSS (@xbrianh ) has plans to deprecate elastic search in the near future, we might have to migrate the queries to JMESPath or something else, I'm not sure how much work needs to be done during that process 🤔

maniarathi:
> It looks good to me.
> 
> My only concern is given we know DSS (@xbrianh ) has plans to deprecate elastic search in the near future, we might have to migrate the queries to JMESPath or something else, I'm not sure how much work needs to be done during that process 🤔

Actually the JMESPath query will be waaaaayyy easier than this query and so should be a super easy transition :) I will be happy to help out at that time as well!

samanehsan (approved):

subscription/elasticsearch_queries/scripts/create_dss_subscription.py
hannes-ucsc: Tentatively, I think only restricting on major version is sensible. Adding entities or properties should not break metadata API. Renaming and deleting entities or properties and changing the type of a property could break metadata API. Where is the definition of what constitutes a minor update?

PR #177, opened 2019-08-19
Add data store notifications to pub/sub queue

samanehsan:
### Purpose
https://broadinstitute.atlassian.net/jira/software/projects/GH/boards/530?selectedIssue=GH-470

Update Lira to add DSS notifications to a google pub/sub topic to fix the timeout/duplicate notifications issue.

### Changes
- Add a new endpoint to Lira that subscribes to the pub/sub topic that contains notifications, and then submits an on-hold workflow to Cromwell.

### Review Instructions
Note: This is only set up for the `broad-dsde-mint-dev` project for testing at the moment. The topic is "hca-notifications-dev" and the subscription is "test-push-subscription"

Using the secondary-analysis/tests/integration_test/run_int_test_locally.sh script, it is possible to test the latest commit on this branch in the "dev" Lira environment.

### Set Up Requirements
For each environment:

- [x] Create a pub/sub topic "hca-notifications-{env}"
- [x] Give the caas service account IAM permissions to publish messages 
- [ ] Verify the domain ownership of "{env}.data.humancellatlas.org" (required for creating subscriptions)
- [ ] Create a subscription with the acknowledgement timeout set to 180s (ensure the pub/sub service account has service account token creator role)
- [x] Change GKE load balancer timeout to 180s 

Note: Set-up is complete for dev, integration and staging

### DIFF
--- a/lira/api/notifications.py
+++ b/lira/api/notifications.py
@@ -15,9 +15,8 @@ logger = logging.getLogger("{module_path}".format(module_path=__name__))
 publisher = pubsub_v1.PublisherClient()
 
 
-# Get notifications from data store
 def post(body):
-    """Process notification and launch workflow in Cromwell"""
+    """Receive notifications from the HCA Data Storage Service and add to a Google pub/sub topic."""
     # Check authentication
     lira_config = current_app.config
     if not lira_utils.is_authenticated(connexion.request, lira_config):
@@ -38,23 +37,24 @@ def post(body):
     future = publisher.publish(
         topic_path, message, origin='lira-dev'
     )
-    logger.info(future.result())
-    return 'ok'
+    message_id = future.result()
+    logger.info(f"Message {message_id} added to topic {topic_name}")
+    return lira_utils.response_with_server_header({"id": message_id}, 200)
 
 
-# Get messages from google pub/sub topic:
 def receive_messages():
+    """Receive and process messages from Google pub/sub topic."""
     # if (request.args.get('token', '') !=
     #         current_app.config['PUBSUB_VERIFICATION_TOKEN']):
     #     return 'Invalid request', 400
-
     envelope = json.loads(request.data.decode('utf-8'))
     data = base64.b64decode(envelope['message']['data'])
     logger.info(data)
-    # submit_workflows(data)
+    # response = submit_workflow(data)
+    # return response
 
 
-def submit_workflows(body):
+def submit_workflow(body):
     lira_config = current_app.config
     uuid, version, subscription_id = lira_utils.extract_uuid_version_subscription_id(
         body
--- a/lira/api/notifications.py
+++ b/lira/api/notifications.py
@@ -21,8 +21,8 @@ def post(body):
     if not lira_utils.is_authenticated(connexion.request, lira_config):
         raise connexion.ProblemException(
             status=401,
-            title="Unauthorized",
-            detail="Authentication failed",
+            title='Unauthorized',
+            detail='Authentication failed',
             headers=lira_utils.LIRA_SERVER_HEADER,
         )
 
@@ -49,8 +49,8 @@ def receive_messages(body):
     if not lira_utils._is_authenticated_pubsub(connexion.request):
         raise connexion.ProblemException(
             status=401,
-            title="Unauthorized",
-            detail="Authentication failed",
+            title='Unauthorized',
+            detail='Authentication failed',
             headers=lira_utils.LIRA_SERVER_HEADER,
         )
     message = body["message"]
@@ -62,7 +62,7 @@ def receive_messages(body):
 def submit_workflow(message):
     """ Process messages and submit on-hold workflow in Cromwell."""
     lira_config = current_app.config
-    data = base64.b64decode(message["data"])
+    data = base64.b64decode(message['data'])
     body = json.loads(data)
     uuid, version, subscription_id = lira_utils.extract_uuid_version_subscription_id(
         body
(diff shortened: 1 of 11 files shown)


lira/api/notifications.py
```diff
@@ -2,32 +2,66 @@
...
+            detail="Authentication failed",
             headers=lira_utils.LIRA_SERVER_HEADER,
         )
 
     logger.info("Notification received")
     logger.info(f"Received notification body: {body}")
 
-    # Get bundle uuid, version and subscription_id
+    project_id = lira_config.google_project
+    topic_name = "hca-notifications-dev"
```
samanehsan: I created a PR to add this field to the Lira config but have not merged it yet: https://github.com/HumanCellAtlas/secondary-analysis-deploy/pull/222

rexwangcc: Sounds good, please don't forget to remove these lines before merging! :octocat:

lira/api/notifications.py
```diff
@@ -2,32 +2,66 @@
...
+    publisher = pubsub_v1.PublisherClient.from_service_account_file(
+        lira_config.caas_key
+    )
+    topic_path = publisher.topic_path(project_id, topic_name)
+    message = json.dumps(body).encode("utf-8")
+    future = publisher.publish(topic_path, message, origin="lira-dev")
```
rexwangcc: parameterize `lira-{}` before merging the PR :octocat:

lira/api/notifications.py
```diff
@@ -2,32 +2,66 @@
...
     logger.info(f"Received notification body: {body}")
 
-    # Get bundle uuid, version and subscription_id
+    project_id = lira_config.google_project
+    topic_name = "hca-notifications-dev"
+    publisher = pubsub_v1.PublisherClient.from_service_account_file(
+        lira_config.caas_key
+    )
+    topic_path = publisher.topic_path(project_id, topic_name)
+    message = json.dumps(body).encode("utf-8")
```
rexwangcc, line 38: I guess `bytes` is a hard-limit here?

lira/api/notifications.py
rexwangcc: Also, why call it `future` here? Does that mean it's indeed a Python concurrent `Future` object? If so, the `future.result()` might need an explicit timeout, otherwise this could take forever and stuck silently when there's something happened  on the google side

lira/lira_utils.py
```diff
@@ -19,6 +21,26 @@
...
+        # Verify and decode the JWT. `verify_oauth2_token` verifies
+        # the JWT signature, the `aud` claim, and the `exp` claim.
+        claim = id_token.verify_oauth2_token(token, google_transport_requests.Request())
+        # Must also verify the `iss` claim.
+        if claim['iss'] not in ['accounts.google.com', 'https://accounts.google.com']:
+            logger.error('Wrong issuer.')
+            return False
+        return True
+    except Exception as e:
+        logger.error('Invalid token: {}\n'.format(e))
```
rexwangcc: nit:
```suggestion
        logger.error(f'Invalid token: {e}\n')
```
for consistency

lira/lira_api.yml
```diff
@@ -84,6 +85,34 @@ paths:
           description: "Unexpected Error"
           schema:
             $ref: '#/definitions/Error'
+  /submissions:
```
rexwangcc, line 88: Let's chat offline about this!

rexwangcc (commented):
Looks good, amazing for implementing this so fast! ⚡️

lira/lira_api.yml
rexwangcc: per offline discussion, pub/sub requires an HTTPS endpoint for `push` model, so this is required. Further discussion will be needed about which endpoint the notifier will hit.

rexwangcc:
Besides the above questions, are we going to rely on Falcon's de-dup ability for now, until we implement the de-dup in a more persistent queue in the future?

lira/api/notifications.py
samanehsan: That's a good point! This will return a future: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/pubsub/cloud-client/publisher.py#L183

lira/api/notifications.py
samanehsan: Yes, I can't find it in the documentation but all of the examples say that the data must be a byte string: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/pubsub/cloud-client/publisher.py#L185

samanehsan:
> Besides the above questions, are we going to rely on Falcon's de-dup ability for now, until we implement the de-dup in a more persistent queue in the future?

I have a second PR (still in-progress) where Lira queries Cromwell for workflows with the input hash before submitting a new one to Cromwell: https://github.com/HumanCellAtlas/lira/pull/178

And then I think we can replace this with de-dup based on a persistent queue.

rexwangcc:
@samanehsan Sounds great! (we might need to look into the persistent queue soon since the querying hash approach won't resolve the issue that pub/sub could send same notifications to all of the 3 lira pods)

lira/lira_utils.py
```diff
@@ -19,6 +21,26 @@
...
+    try:
+        # Get the Cloud Pub/Sub-generated JWT in the "Authorization" header.
+        bearer_token = request.headers.get('Authorization')
+        token = bearer_token.split(' ')[1]
+
+        # Verify and decode the JWT. `verify_oauth2_token` verifies
+        # the JWT signature, the `aud` claim, and the `exp` claim.
+        claim = id_token.verify_oauth2_token(token, google_transport_requests.Request())
+        # Must also verify the `iss` claim.
+        if claim['iss'] not in ['accounts.google.com', 'https://accounts.google.com']:
```
samanehsan, line 36: There is also the option to add an "audience" for the subscription that will get included in the token. Do you know if it makes sense to use the same audience from the DCPAuthClient? @rexwangcc

lira/api/notifications.py
```diff
@@ -37,7 +37,7 @@ def post(body):
     topic_path = publisher.topic_path(project_id, topic_name)
     message = json.dumps(body).encode("utf-8")
     future = publisher.publish(topic_path, message, origin=f"lira-{lira_config.env}")
-    message_id = future.result()
+    message_id = future.result(timeout=60)
```
rexwangcc: would you mind add a comment about this timeout? Theoretically, this should be close or same as the DSS notifier timeout? 🤔

samanehsan:
@rexwangcc that's true! We can add the labels in Lira and keep the de-dup logic in Falcon in the mean time. That way if Lira gets duplicate messages from pub/sub and submits multiple workfows, Falcon will abort the extra ones. My original concern was having a lot of aborted workflows pollute the job manager view from updates to large submissions

lira/api/notifications.py
samanehsan: Sure! Isn't the timeout 60 seconds?

rexwangcc: it is 60 seconds !

lira/lira_utils.py
rexwangcc: I'm not sure actually, shall we ask Trent on this? @Bento007

rexwangcc (approved):
Looks good!

PR #178, opened 2019-08-20
Add input hash label to workflows

samanehsan:
### Purpose
Add input hash labeling back into Lira

### Changes
Add a hash of bundle-specific workflow inputs as a label on the Cromwell workflow. Falcon will use this label to determine whether to start a workflow.

### Review Instructions

- No instructions.

### DIFF
--- a/lira/lira_api.yml
+++ b/lira/lira_api.yml
@@ -254,7 +254,6 @@ definitions:
             example: "v1.0.0"
           lira_version:
             type: string
-            pattern: '(v)?([0-9]+)\.([0-9]+)\.([0-9]+)'
             example: "v1.0.0"
           submit_wdl_version:
             type: string
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ python:
 
 install: "pip install -r requirements.txt"
 
+before_install:
+  - pip install --upgrade pip
+  - pip install --upgrade setuptools
+
 jobs:
   include:
     # The linting test is fast and cheap so run it first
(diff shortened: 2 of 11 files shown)


samanehsan:
I deployed this change to dev and noticed an issue while testing. If a workflow is submitted to Cromwell on-hold, and a second workflow with the same hash-id is also in the queue Falcon will abort both of them. This can be seen here: https://job-manager.caas-prod.broadinstitute.org/jobs?q=bundle-uuid%3Dd727a76f-56b6-4148-b8c1-0ee0d0f02ef9

If Falcon is about to start a workflow but it finds others with the same hash-id, it could compare the submission time and abort it if it is a newer workflow

lira/lira_api.yml
```diff
@@ -254,7 +254,6 @@ definitions:
             example: "v1.0.0"
           lira_version:
             type: string
-            pattern: '(v)?([0-9]+)\.([0-9]+)\.([0-9]+)'
```
samanehsan, line 257: Will add this back in after testing Lira

samanehsan:
Btw, pub/sub keeps sending duplicate messages to Lira, which is why we see two workflows started for a single notification 🤔

samanehsan:
Update: The issue was that both the GKE load balancer and gunicorn request timeouts were set to 60s causing the pub/sub messages to timeout before the timeout set by the subscription. Now, all three timeout limits are 180s

samanehsan:
@rexwangcc does this look good to merge into the main feature branch `se-add-notifications-to-queue`?

rexwangcc (approved):
LGTM!

PR #179, opened 2019-08-27
Fix submissions yml definition

samanehsan:
### Purpose
Include 200 as a valid response code from the "submissions" endpoint in addition to 201.


rexwangcc (approved):
LGTM!!!

PR #181, opened 2019-09-03
Add config flag for testing

samanehsan:
### Purpose
Add a config variable that can be enabled by the mintegration tests to bypass checking that messages are coming from google pub/sub, since it tests lira by running it locally.


charlesdyan (approved):
lgtm

PR #183, opened 2019-09-16
Fix default subscription type

samanehsan:
### Purpose
We are still using elasticsearch subscriptions, so the default subscription type in the scripts to get, create and delete subscriptions should be "elasticsearch" instead of "jmespath"


charlesdyan (approved):
lgtm

PR #184, opened 2019-09-26
Only get human data with 10x-query

samanehsan:
### Purpose
Reverting until there is a way to distinguish "beta" vs "standard" analysis for HCA (or the analysis with mouse reference is approved as a standard analysis).

### Changes

- No changes.

### Review Instructions

- No instructions.


pshapiro4broad (approved):
Looks OK.

PR #186, opened 2019-10-07
Update subscription queries

samanehsan:
### Purpose
Now that the SmartSeq2 and Optimus pipelines are approved to analyze mouse data, update the subscription queries to match on either human or mouse data.

### Changes
- Update the subscription queries to match on either human or mouse data
- Remove "should" clause in Lira: https://github.com/HumanCellAtlas/lira/pull/163

### Review Instructions
- No instructions.

Note: The SS2 query should not be replaced until the mouse reference is added to pipeline-tools: https://github.com/HumanCellAtlas/pipeline-tools/pull/170

### DIFF
--- a/subscription/elasticsearch_queries/10x-query.json
+++ b/subscription/elasticsearch_queries/10x-query.json
@@ -18,8 +18,19 @@
           }
         },
         {
-          "match": {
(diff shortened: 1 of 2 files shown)


rexwangcc (approved):
LGTM! Note we might want to coordinate on when to deploy this change to prod!

PR #188, opened 2019-10-09
Pass pipeline version to workflow inputs

samanehsan:
### Purpose
Getting the pipeline version from the outputs of the analysis workflow requires keeping that hard-coded version variable up to date in the analysis pipeline. This is prone to manual errors and complicates the release process for pipelines in Skylab.

### Changes
Since Lira already knows the pipeline version, pass it into the workflow inputs.

### Review Instructions
- No instructions.

### DIFF
--- a/lira/lira_utils.py
+++ b/lira/lira_utils.py
@@ -217,7 +217,7 @@ def compose_inputs(workflow_name, workflow_version, uuid, version, lira_config):
         workflow_name + '.schema_url': lira_config.schema_url,
         workflow_name + '.cromwell_url': lira_config.cromwell_url,
         workflow_name + '.timestamp': get_utcnow_timestamp(),
-        workflow_name + '.workflow_version': workflow_version,
+        workflow_name + '.pipeline_version': workflow_version,
     }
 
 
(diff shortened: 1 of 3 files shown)


rexwangcc (approved):

lira/lira_utils.py
```diff
@@ -216,6 +217,7 @@ def compose_inputs(workflow_name, uuid, version, lira_config):
         workflow_name + '.schema_url': lira_config.schema_url,
         workflow_name + '.cromwell_url': lira_config.cromwell_url,
         workflow_name + '.timestamp': get_utcnow_timestamp(),
+        workflow_name + '.workflow_version': workflow_version,
```
tbl3rd: Why isn't workflow_name just another member? `{"workflow_name": workflow_name}`?

tbl3rd (approved):

lira/lira_utils.py
samanehsan: We're not using the `workflow_name` as an input to the adapter pipeline so it's not included here. Unless you mean the "workflow_name" prefix isn't necessary if you include `{"workflow_name": workflow_name}`?

PR #190, opened 2019-10-28
Fix monitoring image setup

samanehsan:
### Purpose
Fix set-up for recording workflow metadata in BigQuery.

### Changes
Add `final_workflow_log_dir` to workflow options -- writing to this directory will trigger the Google cloud function that records metadata when each workflow completes: 
https://github.com/broadinstitute/cromwell-task-monitor-bq#metadata-upload

### Review Instructions
- No instructions.


rexwangcc (approved):
The change itself looks good to me. However, from what I've heard, there are a lot of issues with this specific field, reported by other people, so we might want to keep a close eye on this!

PR #194, opened 2019-11-21
Update 10x subscription query

samanehsan:
### Purpose
- Get notified of both 10x v2 & v3 data
- Handle both specific and general library construction ontology terms

### Changes
Match bundles that are either:
- "10X 3' v2 sequencing"
- "10X v2 sequencing" with a "3 prime tag" end bias
- "10X 3' v3 sequencing"
- "10X v3 sequencing" with a "3 prime tag" end bias

### Review Instructions
- No instructions.

### DIFF
--- a/subscription/elasticsearch_queries/10x-query.json
+++ b/subscription/elasticsearch_queries/10x-query.json
@@ -3,13 +3,30 @@
     "bool": {
       "must": [
         {
-          "match": {
-            "files.library_preparation_protocol_json.library_construction_method.ontology_label": "10X v2 sequencing"
-          }
-        },
-        {
-          "match": {
-            "files.library_preparation_protocol_json.end_bias": "3 prime tag"
+          "bool": {
+            "should": [
+              {
+                "match": {
+                  "files.library_preparation_protocol_json.library_construction_method.ontology_label": "10X 3' v2 sequencing"
+                }
+              },
+              {
+                "bool": {
+                  "must": [
+                    {
+                      "match": {
+                        "files.library_preparation_protocol_json.library_construction_method.ontology_label": "10X v2 sequencing"
+                      }
+                    },
+                    {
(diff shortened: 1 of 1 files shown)


subscription/elasticsearch_queries/10x-query.json
```diff
@@ -3,13 +3,30 @@
...
-          }
-        },
-        {
-          "match": {
-            "files.library_preparation_protocol_json.end_bias": "3 prime tag"
+          "bool": {
+            "should": [
+              {
+                "match": {
+                  "files.library_preparation_protocol_json.library_construction_method.ontology_label": "10X 3' v2 sequencing"
```
ESapenaVentura: Instead of `ontology_label`, can we try to match: "files.library_preparation_protocol_json.library_construction_method.ontology": "EFO:0009899"

? Thanks!

subscription/elasticsearch_queries/10x-query.json
```diff
@@ -3,13 +3,30 @@
...
+                }
+              },
+              {
+                "bool": {
+                  "must": [
+                    {
+                      "match": {
+                        "files.library_preparation_protocol_json.library_construction_method.ontology_label": "10X v2 sequencing"
```
ESapenaVentura: Same as above:

"files.library_preparation_protocol_json.library_construction_method.ontology": "EFO:0009310"

ESapenaVentura (changes requested):
Sorry for the intromission but it is a small change and it will be a step towards a better use of the ontologies.

Thanks!

subscription/elasticsearch_queries/10x-query.json
samanehsan: Yes! Thanks for pointing that out!

rexwangcc (approved):

PR #195, opened 2019-11-25
Update pubsub configuration

samanehsan:
### Purpose
When using the notifier script to manually send notifications to Lira, the response time for a single request can take up to 10 seconds, which makes starting analysis of a dataset take significantly longer 🙀 The only task that is being performed at that stage is publishing the message to the pubsub topic, so the configuration should be changed to make this faster.

### Changes
Set the batch message number to 1, so that messages are immediately published to the topic (instead of waiting for multiple messages to send as a batch).

See "Batching to Balance Latency" documentation on publisher client: https://cloud.google.com/pubsub/docs/publisher

### Review Instructions
- No instructions.


rexwangcc (approved):
WOW, this is fantastic!

PR #197, opened 2019-12-04
Fix pubsub latency

samanehsan:
### Purpose
It currently takes Lira anywhere from 5-20 seconds to publish a message to the google PubSub topic. After further investigation, the `google-cloud-pubsub` [python library](https://github.com/googleapis/google-cloud-python/blob/master/pubsub/google/cloud/pubsub_v1/publisher/client.py#L21) is using the `grpc` library which is not compatible with the `gevent` workers by default. However, `grpc` provides a method that will make it compatible.

Related links:
- https://medium.com/@crueda/gunicorn-flask-with-a-side-of-grpc-98a74c9285a7
- https://github.com/grpc/grpc/issues/4629

### Changes
Call `grpc.experimental.gevent.init_gevent()` to patch gRPC so that it is compatible with gevent.

Testing this branch locally, you can see that the time to publish a message drastically decreases 😮 
```
Time to publish:0.0009891986846923828
INFO:lira.api.notifications:Time to return future: 0.44507527351379395
```

### Review Instructions
I don't think any other parts of Lira are using gRPC, but it would be helpful to think through any side-effects this could have.

### DIFF
--- a/lira/api/notifications.py
+++ b/lira/api/notifications.py
@@ -34,7 +34,7 @@ def post(body):
     project_id = lira_config.google_project
     topic_name = lira_config.google_pubsub_topic
 
-    batch_settings = pubsub_v1.types.BatchSettings(max_messages=1, max_latency=0)
+    batch_settings = pubsub_v1.types.BatchSettings(max_messages=1)
     publisher = pubsub_v1.PublisherClient.from_service_account_file(
         lira_config.caas_key, batch_settings=batch_settings
     )
--- a/lira/api/notifications.py
+++ b/lira/api/notifications.py
@@ -1,4 +1,5 @@
 import connexion
+import gevent
 import json
 import logging
 import io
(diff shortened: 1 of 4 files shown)


lira/lira.py
```diff
@@ -64,6 +66,12 @@
     'lira_api.yml', resolver=resolver, validate_responses=True, arguments=arguments
 )
 
+# Patch GRPC (used by Google PubSub) to be gevent-compatible
```
rexwangcc: optional: It would be helpful to also append the link to the issue: https://github.com/grpc/grpc/issues/4629 so we don't forget the detailed context of this issue.

lira/api/notifications.py
```diff
@@ -37,10 +39,16 @@ def post(body):
     )
     topic_path = publisher.topic_path(project_id, topic_name)
     message = json.dumps(body).encode("utf-8")
+    start = time.time()
```
rexwangcc: Reminder to remove these time measurements before merging the PR :octocat:

rexwangcc (approved):
LGTM! As of side effects, I can't think of any possible downsides of this patch since it's conditional. Great work on fixing this!!!! Thanks and yay 🎉

